PR #3, opened 2022-11-30
Add interfaces

davesnx:


lib/react.mli
```diff
@@ -0,0 +1,146 @@
+[@@@warning "-67"] (* There's an unused parameter on functor Fragment *)
```
anmonteiro, line 1: you could probably add it to the type itself 

```ocaml
type x = ... [@@warning "-67"]
```

davesnx: Totally, I don't understand the error since complains about a functor where I don't have one.

Thanks for sneaking on the Pars :P

PR #29, opened 2023-09-16
Add flake + Context fix

davesnx:


packages/react/cross-test/js/dune
```diff
@@ -0,0 +1,18 @@
...
+; this simply builds the melange version of the same code tested in the native folder to ensure 
+; that the interfaces of server-reason-react line up with reason-react
+(melange.emit
+  (target main)
+  (alias js)
+  (module_systems es6)
+  (modules)
+  (libraries cross_test_js))
+
+(copy_files# "../native/shared/test_context.re")
```
davesnx, line 18: Can you move this under demo/?

packages/react/src/react.mli
```diff
@@ -93,10 +93,11 @@ type 'a context = {
 }
 
 module Context : sig
-  val provider : 'a context -> 'a provider
+  type 'a t
```
davesnx, line 96: Nice, forgot this

packages/react/cross-test/js/dune
joprice: I was thinking it should be in the test context so it fails when running dune build. I just tried introducing a breaking change in the demo and it didn't cause the build to fail. Maybe if I run with a specific target it would?

joprice: I tried to repeat that and it did fail this time. Not sure why it didn't before. I can move it to the demo.

PR #32, opened 2023-09-16
Create a demo with copy_files

davesnx:


demo/shared/native/lib/MelRaw.re
```diff
@@ -0,0 +1,36 @@
+let%browser_only initWebsoket = [%mel.raw
```
joprice, line 1: typo `initWebsoket -> initWebsocket`

PR #35, opened 2023-09-16
add opam cache to gh actions

joprice:


joprice:
This doesn't seem to improve much and the setup-ocaml action already seems to have some sort of caching. Nevertheless, it does 7 minutes of work on each run.

davesnx:
Merging since it improves, but really not satisfied with opam + gh actions

PR #54, opened 2023-09-29
reorder style props

joprice:
This reorders the styles to match the order used by reason-react, in order to silence React warnings about inline styles on the client not matching the server's values.

The list was generated from the reason-react repo with the following script:

```sh
for style in $(cat src/dOMStyle.re | grep "~" | sed 's/.*\~\([^:]*\):.*/\1/'); do
  dashed=$(echo "$style" | sed 's/\([A-Z]\)/-\L\1/g')
  echo "|> add \"$dashed\" $style"
done
```


davesnx (approved):
Super good. Can you add the script and the comment in the file directly?

PR #65, opened 2023-10-02
add tag to errors to ease debugging

joprice:
This adds a tag to the errors in generated code to make it easier to debug what part of the ppx they came from. Maybe having this in the final code is a bit odd, so it could be hidden behind a new ppx flag for debugging, or shortened to an integer representing each error context.

### DIFF
--- a/packages/ppx/browser.ml
+++ b/packages/ppx/browser.ml
@@ -36,6 +36,11 @@ module Effect = struct
 end
 
 module Browser_only = struct
+  let impossible ~label loc message =
+    let message = "[" ^ label ^ "]" ^ ": " ^ message in
+    let message = Builder.estring ~loc message in
+    [%expr raise (ReactDOM.Impossible_in_ssr [%e message])]
+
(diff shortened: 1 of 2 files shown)


packages/ppx/browser.ml
```diff
@@ -36,6 +36,11 @@ module Effect = struct
 end
 
 module Browser_only = struct
+  let impossible ~label loc message =
+    let message = "[" ^ label ^ "]" ^ ": " ^ message in
```
davesnx, line 40: Maybe instead of the label we should render the location directly?

davesnx: I'm afraid that those labels are very Parsetree specific and not user-centric

joprice: Good idea. Using the location could also reduce code size by avoiding copying large strings of the function definition.

davesnx:
Closing in favor of https://github.com/ml-in-barcelona/server-reason-react/pull/70

PR #68, opened 2023-10-09
Melange native ppx

jchavarri:
Adds a ppx to be used in native builds, that transforms all externals using `mel.` attrs into functions that raise.

The ppx preserves the type annotations from the external. I tried to produce `let foo: bar` rather than `let (foo: bar)` but it seems the version of the ast we are using (5.0) doesn't include yet the `pvb_constraint` field which is the one used for that. But shouldn't be an issue.

The only remaining thing to discuss is how to handle Melange-special externals where the type is not of kind `Ptyp_arrow`, e.g. `external foo: t = "foo"`. Right now, the ppx will generate a `let` binding that just raises, but this is probably not what we want to do. Or is it?

### DIFF
--- a/packages/ppx/dune
+++ b/packages/ppx/dune
@@ -33,3 +33,13 @@
  (preprocess
   (pps ppxlib.metaquot))
  (kind ppx_rewriter))
+
+(library
+ (name melange_native_ppx)
+ (modules melange_native)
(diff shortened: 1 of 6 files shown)


davesnx:
Merging since I want to try it out, yolo

PR #107, opened 2023-12-01
add test for spellcheck

jchavarri:
The `spellCheck` attribute is handled differently if it goes through ppx vs set manually. The PR adds a test to show it.


packages/server-reason-react-ppx/test/test.re
```diff
@@ -265,6 +265,20 @@ let svg_2 = () => {
...
+  assert_string(
+    ReactDOM.renderToStaticMarkup(div),
+    {|<div spellcheck="false"></div>|},
+  );
+};
+
+let spellCheck_without_ppx = () => {
+  let div =
+    React.createElement("div", ReactDOM.domProps(~spellCheck=false, ()), []);
+  assert_string(ReactDOM.renderToStaticMarkup(div), "<div ></div>");
```
jchavarri, line 279: Notice how in this case, the attribute disappears, when it should be like the one above.

davesnx: spellCheck is a special case, and is not handled by server-reason-react.

Boolean attributes that are set to false aren't rendered on the resultant HTML, in this case spellCheck is an Enumerated attribute (I learn that today https://developer.mozilla.org/en-US/docs/Glossary/Enumerated)

I should add Enumerated as a Attribute variant.

jchavarri:
After looking more at the code, I am realizing there is something going on in the current code that conflates both types of props. `ReactDOM.createDOMElementVariadic` takes in a list of `React.JSX.prop`:

https://github.com/ml-in-barcelona/server-reason-react/blob/44bb2218ca041c319e982dbd3cf44560c7e90a68/packages/reactDom/src/reactDOM.mli#L37

This is also the type that is returned by `domProps` function:

https://github.com/ml-in-barcelona/server-reason-react/blob/44bb2218ca041c319e982dbd3cf44560c7e90a68/packages/reactDom/src/reactDOM.mli#L532

The above allows to define `createDOMElementVariadic` in terms of `React.createElement`:

https://github.com/ml-in-barcelona/server-reason-react/blob/44bb2218ca041c319e982dbd3cf44560c7e90a68/packages/reactDom/src/reactDOM.ml#L234-L236

I think that the mentioned `ReactDOM` functions should work with `DomProps.props` or a different type, and not with the more generic `React.JSX.prop`. Does that make sense?

davesnx:
- `React.JSX.props` are the values set at runtime 
- while `DomProps.props` are the types used by the ppx to add annotations

I'm unsure how to unify both types (It's probably doable)

packages/reactDom/src/reactDOM.ml
```diff
@@ -240,6 +240,9 @@ let add kind value map =
 
 type dangerouslySetInnerHTML = { __html : string } [@@boxed]
 
+(* `Booleanish_string` are JSX attributes represented as boolean values but rendered as strings on HTML https://github.com/facebook/react/blob/a17467e7e2cd8947c595d1834889b5d184459f12/packages/react-dom-bindings/src/server/ReactFizzConfigDOM.js#L1165-L1176 *)
+let booealish_string name v = JSX.string name (string_of_bool v)
```
jchavarri, line 244:
```diff
- let booealish_string name v = JSX.string name (string_of_bool v)
+ let booleanish_string name v = JSX.string name (string_of_bool v)
```

jchavarri (commented):
thx

packages/reactDom/src/reactDOM.ml
davesnx: Fixed 526880c80fc54dc822de640efe4073d9d2fc3f85

PR #109, opened 2023-12-07
Support melange v3

jchavarri:
Melange v3 is changing a lot of the APIs for libs like `melange.js`. This PR is catching up with the changes in server-reason-react native implementations.

TODO

- [x] Js.Array
- [x] Js.Dict
- [x] Js.Date
- [x] Js.Int
- [x] Js.Float
- [x] Js.Re
- [x] Js.Math
- [x] Js.Global
- [x] Js.String
- [x] Remove Js.Option, Js.List

### DIFF
--- a/packages/melange.js/js.ml
+++ b/packages/melange.js/js.ml
@@ -1359,7 +1359,7 @@ module Dict : Dictionary = struct
   let unsafeDeleteKey (dict : 'a t) (key : key) = Hashtbl.remove dict key
 end
 
-module Global = struct
+module Global : sig
   (** Contains functions available in the global scope
     ([window] in a browser context) *)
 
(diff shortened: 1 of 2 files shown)


packages/melange.js/js.ml
```diff
@@ -200,196 +200,254 @@ module Exn = struct
...
-  type 'a t = 'a array
+module Array : sig
   (** JavaScript Array API *)
 
+  type 'a t = 'a array
   type 'a array_like
 
-  let from _ = notImplemented "Js.Array2" "from"
-  let fromMap _ _ = notImplemented "Js.Array2" "fromMap"
+  val from : 'a array_like -> 'a array
```
jchavarri, line 209: I know this means we're duplicating signatures here and in mli. But it's helpful to generate the `mli` from melange original repo module (e.g. generate `Js_array.mli` from `Js_array.ml`), replace `external` with `val` and just paste it here, to make sure both interfaces stay in sync.

packages/melange.js/js.mli
```diff
@@ -286,51 +286,28 @@ module Array : sig
...
-  val lastIndex : t -> int
-  val setLastIndex : t -> int -> unit
-  val exec_ : t -> string -> result null
-  val exec : string -> t -> result null
-  val test_ : t -> string -> bool
-  val test : string -> t -> bool
+  val exec : str:string -> t -> result option
+  val test : str:string -> t -> bool
+
+  val matches : result -> string array
```
davesnx, line 309: Then it shouldn't be part of the interface, right?

davesnx:
Since melange@3 breaks the interface with js module, does it make sense to update the pins?

packages/melange.js/js.mli
jchavarri: But then, how can you use it from native?

PR #124, opened 2024-03-19
Try multi-context setup with dune

jchavarri:
This PR is the first attempt as using the multi-context support for libraries with the same name that is being added in Dune (see related issue [ocaml/dune#10222](https://github.com/ocaml/dune/issues/10222).

The gist of the improvement is that we define two Dune contexts for each "universal" project: an OCaml native context (`default`) and a Melange context (`melange`). Then, all native exes and libraries have to be annotated with `(= %{context_name} "default"))` and all Melange emit and libraries have to be annotated with `(= %{context_name} "melange"))`, so that they get only get built under the proper context.

This approach allows to:
- remove `copy_files` rules from libraries with shared code
- get rid of the multi-folder layout for shared libraries (all modules and dune definitions can be in a single folder)
- use Dune wrapped libraries

Besides this, it should potentially be possible to get the editor definitions from the Melange or native cmts by leveraging the `(merlin)` field in Dune contexts, but something else seems to be broken that I will have to be investigate.

For now draft PR until a version of Dune is eventually released with those improvements.

See related discussion: [melange-re/melange#694](https://github.com/melange-re/melange/discussions/694)


demo/client/dune
```diff
@@ -2,9 +2,11 @@
  (alias client)
  (target app)
  (enabled_if
-  (= %{profile} "dev"))
+  (and
+   (= %{profile} "dev")
+   (= %{context_name} "melange")))
```
anmonteiro, line 7: not a big deal but I think we can use `(= %{context_name} melange)` and becomes even cleaner :chefs_kiss:

jchavarri: nice. updated in https://github.com/ml-in-barcelona/server-reason-react/pull/124/commits/9095b40a920be0d513207a88018e54d786233b05.

jchavarri:
In the last commits, I removed the last usages of `copy_files` in the `url` library. For this, I had to change the base branch to the PR in #127 so that we're able to include both implementations of the `URL` module in the same file.

We could also remove the `wrapped false` now in the `url` library, as both native and melange versions share the name. But this change was adding a lot more noise to the diff so maybe can be done separately.

The only missing parts are:

- wait for Dune to make a release to include the multi-context support
- update docs to remove references to `copy_files`

jchavarri:
Doesn't seem we'll be moving forward with multi-context: https://github.com/ocaml/dune/issues/10507#issuecomment-2189900682

PR #125, opened 2024-03-27
Make useUrl return the provided serverUrl

### LINKED ISSUE #123: Make ReasonReactRouter.useUrl return ~serverUrl or throw
Hi, this is a really awesome library! I'm having fun re-learning OCaml and ReasonML with it.

Based on the documentation for `ReasonReactRouter.useUrl`:
```
/** hook for watching url changes.
 * serverUrl is used for ssr. it allows you to specify the url without relying on browser apis existing/working as expected
 */
```

This is reflected in the implementation where if the `serverUrl` is passed, it just returns that--otherwise, it calls into JS in order to obtain the URL (which is not possible in a server context).

In server-reason-react, `useUrl` returns an empty `url` and completely ignores `serverUrl`. If used in a shared component, routing silently fails and the `[]` gets matched every time:
```reason
module App = {
  [@react.component]
  let make = (~serverUrl: option(ReasonReactRouter.url)=?) => {
    let url = ReasonReactRouter.useUrl(~serverUrl?, ());
    switch (url.path) {
    | [] => <div> {React.string("Index!")} </div>
    | [name] => <Counter name />
    | _ => <div> {React.string("Not Found!")} </div>
    };
  };
};
```

The workaround I came up with is:
```reason
module App = {
  [@react.component]
  let make = (~serverUrl: option(ReasonReactRouter.url)=?) => {
    let url =
      switch%platform (Runtime.platform) {
      | Client => ReasonReactRouter.useUrl(~serverUrl?, ())
      | Server =>
        switch (serverUrl) {
        | Some(serverUrl) => serverUrl
        | None => raise(Invalid_argument("Server component needs URL!"))
        }
      };
  
    switch (url.path) {
    | [] => <div> {React.string("Index!")} </div>
    | [name] => <Counter name />
    | _ => <div> {React.string("Not Found!")} </div>
    };
  };
};
```
Alternatively, having separate implementations as suggested by the [code-structure](https://ml-in-barcelona.github.io/server-reason-react/local/server-reason-react/code-structure.html) docs is another reasonable solution.

What do you think of defining `useUrl` as the following?
```reason
let useUrl ?(serverUrl : ReasonReactRouter.url option) (_ : unit) =
  match serverUrl with
  | Some serverUrl -> serverUrl
  | None -> raise (Invalid_argument "Must provide serverUrl")
```

I could also write the PR for it--would be cool as a first contribution to the ecosystem 😄

purefunctor:
Closes #123. This changes `useUrl` to mirror the behaviour of `reason-react` when the `serverUrl` is passed.

See https://github.com/reasonml/reason-react/blob/185d6972306b1009bd3fd3497510b346e757c4e3/src/ReasonReactRouter.re#L191-L197

### DIFF
--- a/packages/react/src/ReasonReactRouter.ml
+++ b/packages/react/src/ReasonReactRouter.ml
@@ -21,4 +21,6 @@ let url ?serverUrlString () =
 let dangerouslyGetInitialUrl = url
 let watchUrl _callback () = ()
 let unwatchUrl _watcherID = ()
-let useUrl ?serverUrl:_ () = url ()
+
+let useUrl ?(serverUrl : url option) () =
+  match serverUrl with Some serverUrl -> serverUrl | None -> url ()


davesnx (approved):
Looks perfect. Thanks @purefunctor

PR #127, opened 2024-04-10
browser-ppx: process stritems

jchavarri:
Allows to do things like:

```ocaml
module T = struct
  type t = Js.Json.t
end [@@platform js]

module U = struct
  type t = string
end [@@platform native]
```

Both `platform` or `browser-ppx` can be used as the attribute tag (this change applies to every existing preprocessing of `browser-ppx` as well).

### DIFF
--- a/packages/browser-ppx/ppx.ml
+++ b/packages/browser-ppx/ppx.ml
@@ -4,6 +4,9 @@ module Builder = Ast_builder.Default
 type target = Native | Js
 
 let mode = ref Native
+let tag = "browser_ppx"
+let alt_tag = "platform"
+let is_platform_tag str = String.equal str tag || String.equal str alt_tag
 
 module Platform = struct
   let pattern = Ast_pattern.(__')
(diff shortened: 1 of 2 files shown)


packages/browser-ppx/ppx.ml
```diff
@@ -273,14 +276,202 @@ module Browser_only = struct
...
+        if should_keep_many cds (fun cd -> cd.pci_attributes) = `keep then
+          Some sigi
+        else None
+    | Psig_class_type ctds ->
+        if should_keep_many ctds (fun ctd -> ctd.pci_attributes) = `keep then
+          Some sigi
+        else None
+    | Psig_extension _ | Psig_attribute _ -> Some sigi
+
+  let preprocess_impl _exp_ctxt str =
```
jchavarri, line 450: I am not sure what the `_exp_ctxt` variable is about, it was introduced in ppxlib v2, @davesnx do you know?

packages/browser-ppx/ppx.ml
```diff
@@ -273,14 +276,202 @@ module Browser_only = struct
...
-  Driver.V2.register_transformation "browser_ppx"
+  Driver.V2.register_transformation tag
     ~rules:
       [
         Browser_only.expression_rule;
         Browser_only.structure_item_rule;
         Platform.rule;
       ]
+    ~preprocess_impl:Preprocess.preprocess_impl
+    ~preprocess_intf:Preprocess.preprocess_intf
```
jchavarri, line 477: Is there a way to do this processing using just `rules` so no `preprocess_impl` or `preprocess_intf` is used?

davesnx: Haven't looked into the code, yet, but would be nice

packages/browser-ppx/tests/preprocess.t
```diff
@@ -0,0 +1,23 @@
...
+  include struct
+    type t = Js.Json.t
+  end [@@platform js]
+
+Without -js flag, it picks the block with `[@@platform native]`
+
+  $ ./standalone.exe -impl input.ml | ocamlformat - --enable-outside-detected-project --impl
+  include struct
+    type t = string
+  end [@@platform native]
```
davesnx, line 23: It should remove the annotation no?

packages/browser-ppx/tests/preprocess.t
```diff
@@ -0,0 +1,23 @@
+  $ cat > input.ml << EOF
+  > include struct
+  >   type t = Js.Json.t
+  > end [@@platform js]
+  > 
+  > include struct
+  >   type t = string
+  > end [@@platform native]
```
davesnx, line 8: wondering what happens (if ppx raises) when you add platform to only one

packages/browser-ppx/ppx.ml
```diff
@@ -273,14 +276,202 @@ module Browser_only = struct
...
+          Some sigi
+        else None
+    | Psig_extension _ | Psig_attribute _ -> Some sigi
+
+  let preprocess_impl _exp_ctxt str =
+    match str with
+    | { pstr_desc = Pstr_attribute attr; _ } :: rest
+      when is_platform_tag attr.attr_name.txt ->
+        if eval_attr attr = `keep then rest else []
+    | _ -> List.filter_map apply_config_on_structure_item str
```
davesnx, line 455: I know you might not implemented this part, but why do we have 2 passes? 

Probably [@platform] can be applied to any structure, not only modules in this case.

davesnx (approved):
Happy to merge it as it is, we can push for a better impl with:

- usage of context free
- have proper errors when one is missing
- more test cases

packages/browser-ppx/tests/preprocess.t
jchavarri: Yes, I think it should. Will look into this.

packages/browser-ppx/ppx.ml
jchavarri: I'm not sure I understand the q, where are the two passes? What I understand this code is doing is:

- If the str item itself has an the platform attr, process it
- Otherwise, process whatever is inside the str item

packages/browser-ppx/tests/preprocess.t
jchavarri: I tried removing the attributes but it's quite cumbersome. The approach I took is to transform `should_keep` into st like this:

```ocaml
  let rec should_keep attrs =
    let rec f acc attrs =
      match attrs with
      | [] -> `keep (List.rev acc)
      | attr :: attrs -> (
          match eval_attr attr with
          | `found_drop -> `drop
          | `found_keep -> f acc attrs
          | `not_found -> f (attr :: acc) attrs)
    in
    f [] attrs
```

The main issue I found is on the `apply_config_on_structure_item` function. Instead of the current simple approach, it becomes this monster:

```ocaml
  let apply_config_on_structure_item stri =
    match stri.pstr_desc with
    | Pstr_typext ({ ptyext_attributes = attrs; _ } as payload) -> (
        match should_keep attrs with
        | `keep attrs ->
            Some (Pstr_typext { payload with ptyext_attributes = attrs })
        | `drop -> None)
    | Pstr_modtype ({ pmtd_attributes = attrs; _ } as payload) -> (
        match should_keep attrs with
        | `keep attrs ->
            Some (Pstr_modtype { payload with pmtd_attributes = attrs })
        | `drop -> None)
    | Pstr_open ({ popen_attributes = attrs; _ } as payload) -> (
        match should_keep attrs with
        | `keep attrs ->
            Some (Pstr_open { payload with popen_attributes = attrs })
        | `drop -> None)
    | Pstr_include ({ pincl_attributes = attrs; _ } as payload) -> (
        match should_keep attrs with
        | `keep attrs ->
            Some (Pstr_include { payload with pincl_attributes = attrs })
        | `drop -> None)
    | Pstr_exception ({ ptyexn_attributes = attrs; _ } as payload) -> (
        match should_keep attrs with
        | `keep attrs ->
            Some (Pstr_exception { payload with ptyexn_attributes = attrs })
        | `drop -> None)
    | Pstr_primitive ({ pval_attributes = attrs; _ } as payload) -> (
        match should_keep attrs with
        | `keep attrs ->
            Some (Pstr_primitive { payload with pval_attributes = attrs })
        | `drop -> None)
    | Pstr_eval (str_desc, attrs) -> (
        match should_keep attrs with
        | `keep attrs -> Some (Pstr_eval (str_desc, attrs))
        | `drop -> None)
    | Pstr_module ({ pmb_attributes = attrs; _ } as payload) -> (
        match should_keep attrs with
        | `keep attrs ->
            Some (Pstr_module { payload with pmb_attributes = attrs })
        | `drop -> None)
      ...
 ```

I am sure there are ways in which it could be simplified, but after discussing with @davesnx we're punting on this for later (maybe we end up using extensions instead of attributes...)

jchavarri:
> usage of context free

I took a stab at this too. It seems non-trivial because context-free:
- only supports extensions, not attributes, so we should change the API somehow to adapt to that
- it expects to return _something_ from the rule function, but we need to erase nodes fully from the ast (Some / None)

I am not sure how to work around that 🤔

PR #129, opened 2024-04-10
Remove data-reactroot attr from ReactDom render_element

pedrobslisboa:
## Motivation

The `data-reactroot` was available before v18 when calling renderToString, but react removed it. 
It can be checked by seeing that renderToString doesn't return a string element with `data-reactroot` at this test:
https://github.com/facebook/react/blob/ed3c65caf042f75fe2fdc2a5e568a9624c6175fb/packages/react-dom/src/__tests__/ReactServerRendering-test.js#L4

The commit which removed that is this one:
https://github.com/facebook/react/commit/940f48b999a3131e77b2545bd7ae252ef27ae6d1

## What this PR does

- Remove `react_root_attr_name` and its follow dependencies;
- Update  renderToString tests.

## Checklist

- [x] Unit tests:

image

### DIFF
--- a/packages/reactDom/src/ReactDOM.ml
+++ b/packages/reactDom/src/ReactDOM.ml
@@ -60,9 +60,6 @@ let attributes_to_string attrs =
   | [] -> ""
   | rest -> " " ^ (rest |> String.concat " " |> String.trim)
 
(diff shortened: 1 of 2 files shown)


davesnx:
Super good @pedrobslisboa

PR #133, opened 2024-05-06
Feat/melange json

pedrobslisboa:
# Why

Create a universal way to handle JSON.

The code was based on melange-json:
- Decode: https://github.com/melange-community/melange-json/blob/55a301aef1f555bb754f9ae8361c786bb60974cd/src/Json_decode.ml
- Encode: https://github.com/melange-community/melange-json/blob/55a301aef1f555bb754f9ae8361c786bb60974cd/src/Json_encode.ml
- Json: https://github.com/melange-community/melange-json/blob/55a301aef1f555bb754f9ae8361c786bb60974cd/src/Json.ml

The idea is to keep it up to date with melange-json since it's a huge community tool.

# How

The Yojson was used to be fast and with reliability a great tool.

## Warning ⚠️ 

The `type t` doesn't exist on melange-json, but I think it should be there then we can have an isomorphic module with t type instead of Js.Json and Yojson.Basic.t. 

Avoid this kind of problem:

```ocaml
(* universal function *) 
let foo value: ??? = Json.Decode.string value
```

It could be:
```ocaml
(* universal function *) 
let foo value: Json.t = Json.Decode.string value
```

Unify the JSON type.

ASAP I will add an issue or pull request there.


davesnx:
This work mutated into https://github.com/pedrobslisboa/universal-json

PR #134, opened 2024-05-17
Support assets in `mel.module`

jchavarri:
This PR adds initial support to compile externals like:

```
external img : string = "default" [@@mel.module "./image.svg"]
```

Into something like:

```
let img = "/image-DWDWZCEH.svg"
```

~The output file naming follows esbuild default shape and hash function. There is no option to configure anything else for now, we can slowly introduce as needed.~ Edit: the ppx now support choosing either `-bundler esbuild` or `-bundler webpack`. The default is webpack.

I decided to vendor the libraries `base32` and `xxhash` for simplicity, to avoid forcing consumers to pin or download new packages, as they are quite small libraries and the versions that were needed are not available in opam.


jchavarri:
I have updated the PR to support both Webpack and Esbuild, via a new flag `-bundler` that can be passed to `melange.ppx`.

For now the generated filenames follow the default name scheme used on each bundler case. We can add more features later on while testing, e.g. file templates like `[name][hash][ext]` like some bundlers do, or the ability to add prefixes to resulting filenames.

I also have left some failed attempt to produce Rspack hashes commented, as currently rspack doesn't support proper content hashes.

I decided to not include tests against the actual Webpack or Esbuild bundlers hashes because these hashes should not change over time, and it would slow down and complicate the test suite quite a bit.

packages/melange.ppx/ppx.ml
```diff
@@ -401,6 +544,20 @@ module Debug = struct
...
+         | "esbuild" -> Mel_module.bundler := Esbuild
+         | _ ->
+             failwith
+               (Printf.sprintf
+                  {|Unknown value %S passed as -bundler flag in melange.ppx, valid values: "webpack", "esbuild"|}
+                  str)))
+    ~doc:
+      "generate paths to assets in mel.module using the file name scheme of \
+       the bundler of choice";
+  Driver.V2.register_transformation ~impl:structure_mapper
```
jchavarri, line 561: Changing to `V2` here is what allows to get the `ctxt` in `structure_mapper`, which is used later on to get the path to the file being processed (so that we can open it to get its content for the hash).

.github/workflows/ci.yml
```diff
@@ -55,6 +55,14 @@ jobs:
           dune-cache: true
           opam-disable-sandboxing: true
 
+      - name: Install xxhash (ubuntu)
+        if: matrix.os == 'ubuntu-latest'
+        run: sudo apt-get install -y libxxhash-dev
```
davesnx, line 60: would it mean that all consumers of server-reason-react need to have this installed?

packages/melange.ppx/ppx.ml
```diff
@@ -256,8 +370,32 @@ let transform_external pval_name pval_attributes pval_loc pval_type =
...
+        | Some str ->
+            (* If it has asset payload (file with extension), calculate hash and replace external *)
+            let name = Builder.pvar ~loc:pval_name.loc pval_name.txt in
+            let path =
+              let asset_path = Filename.(concat (dirname module_path) str) in
+              (* todo: maybe read line by line of buffered for large files *)
+              let ic = open_in asset_path in
+              let n = in_channel_length ic in
+              let s = really_input_string ic n in
+              close_in ic;
```
davesnx, line 387: I used `In_channel.with_open_bin file In_channel.input_all` in the past, if you want to avoid having to open/close_in

packages/melange.ppx/ppx.ml
```diff
@@ -256,8 +370,32 @@ let transform_external pval_name pval_attributes pval_loc pval_type =
...
+      else if Mel_module.has_attr pval_attributes then
+        match Mel_module.asset_payload pval_attributes with
+        | None ->
+            (* If it doesn't have asset payload, we error out as it must be some .js module or package being imported *)
+            [%stri [%%ocaml.error [%e mel_module_found_in_native_message ~loc]]]
+        | Some str ->
+            (* If it has asset payload (file with extension), calculate hash and replace external *)
+            let name = Builder.pvar ~loc:pval_name.loc pval_name.txt in
+            let path =
+              let asset_path = Filename.(concat (dirname module_path) str) in
```
davesnx, line 382: Smart, this way we don't care what dune dones

packages/melange.ppx/ppx.ml
```diff
@@ -256,8 +370,32 @@ let transform_external pval_name pval_attributes pval_loc pval_type =
...
+              let ic = open_in asset_path in
+              let n = in_channel_length ic in
+              let s = really_input_string ic n in
+              close_in ic;
+              let filename_fn =
+                match !Mel_module.bundler with
+                | Webpack -> Mel_module.Webpack.filename
+                | Esbuild -> Mel_module.Esbuild.filename
+              in
+              let prefix = (* todo: read from config *) "/" in
```
davesnx, line 393: What's prefix here? The cdn url/etc?

davesnx (approved):

packages/melange.ppx/ppx.ml
```diff
@@ -401,6 +544,20 @@ module Debug = struct
 end
 
 let () =
-  Driver.register_transformation ~impl:structure_mapper
+  Driver.add_arg "-bundler"
```
davesnx, line 547: at first I didn't like the idea of coupling our melange_ppx with a bundler, and I would prefer to keep a `--assets-hash-fn` where it specifies "xxhash64" or even a format like webpack, but after seeing the complexity it carries... It's perfectly fine to keep it coupled.

Did you thought the same?

.github/workflows/ci.yml
jchavarri: Yes :( Alternatives can be:
- put the melange.ppx in a separate package
- reimplement the xxhash lib in OCaml
- ...

packages/melange.ppx/ppx.ml
jchavarri: Yeah anything that comes before the file name itself.

.github/workflows/ci.yml
davesnx: and what about adding libxxhash-dev depext?

packages/melange.ppx/ppx.ml
jchavarri: It's a bit tricky, because the behavior changes a bit between bundlers:

- webpack is the most flexible of those i've investigated, allowing to use [any kind of hash function](https://webpack.js.org/configuration/output/#outputhashfunction) and [any kind of template](https://webpack.js.org/guides/asset-modules/#custom-output-filename) to build the resulting asset name
- rspack allows [any kind of template](https://www.rspack.dev/config/output#outputassetmodulefilename), but hash functions are delimited to xxhash64 or md4 (and the hashes are not really from the file content, unlike webpack)
- esbuild [allows any template](https://esbuild.github.io/api/#asset-names), but only supports xxhash64 (postprocessed with base32, which makes it different from the two above)

How do you think we should model this? It certainly seems we should allow users to configure somewhere the file name template to be used to generate the final paths, right? But we should somehow guarantee that the templates are also compatible across bundlers?

In my mind, this wasn't blocking an initial mvp of this feature because users can always update their JS bundler config to use the default behavior of each bundler (which is what is currently implemented). Wdyt?

PR #142, opened 2024-06-27
Add `Float.toFixed` and friends

jchavarri:
Previously based on `asset-prefix` (#140).

Adds some functions to `Js.Float`.


jchavarri:
I'm not convinced this is the way to go. Because the behavior in native will differ from the JS one in some cases, it kind of defeats one of the main purpose of server-reason-react (render same HTML in server with OCaml and client with Melange).

I think it's better to keep current `not_implemented` alerts and let users use alternatives from the stdlib, which guarantees the behavior is the same.

@davesnx feel free to reopen if anything's recoverable :)

davesnx:
Currently the `Js` module is marked as very unstable, so keeping a half-backed implementation sounds good to merge to me.

PR #161, opened 2024-08-11
Add traverse to handle nested module with platform anotation

pedrobslisboa:
## Description

The nested module does not work with the platform attribute.

```ocaml
module X = struct
  include struct
    type t = Js.Json.t
    let a = 2 + 2
  end [@@platform js]
   
  include struct
    type t = string
    let a = 4 + 4
  end [@@platform native]
end

(* With -js flag output: *)
  module X = struct
    include struct
      type t = Js.Json.t
  
      let a = 2 + 2
    end [@@platform js]
  
    include struct
      type t = string
  
      let a = 4 + 4
    end [@@platform native]
  end

(* With no flag output: *)
  module X = struct
    include struct
      type t = Js.Json.t
  
      let a = 2 + 2
    end [@@platform js]
  
    include struct
      type t = string
  
      let a = 4 + 4
    end [@@platform native]
  end
```

As you can see, it does not work well if we want nested modules with the platform attribute.

## How

We were using a standard function to analyze the TopLevel structure items when we could use a Traverse map that, according to the PPXLib doc:

> A traversal is a recursive function that will be called on a value and recursively on all its sub-values, combining the result in a certain way.

Using traverse, it looks for all structures recursively, and the output of the description sample would be:

```ocaml
(* With -js flag output: *)
module X = struct
  include struct
    type t = Js.Json.t
    let a = 2 + 2
  end [@@platform js]
end

(* Without -js flag output: *)
module X = struct
  include struct
    type t = string
    let a = 4 + 4
  end [@@platform native]
end
```

@davesnx I know that we talked about it, and you said that maybe Global Transformation was not the way, but I was studying more about `prepocess_impl` and `impl`, and found the `Ast_traverse`, which worked on it. Does it make sense for now?

### DIFF
--- a/packages/browser-ppx/ppx.ml
+++ b/packages/browser-ppx/ppx.ml
@@ -510,19 +510,26 @@ module Preprocess = struct
         else None
     | Psig_extension _ | Psig_attribute _ -> Some sigi
 
-  let preprocess_impl _exp_ctxt str =
+  let preprocess_impl str =
     match str with
     | { pstr_desc = Pstr_attribute attr; _ } :: rest
       when is_platform_tag attr.attr_name.txt ->
         if eval_attr attr = `keep then rest else []
     | _ -> List.filter_map apply_config_on_structure_item str
 
-  let preprocess_intf _exp_ctxt sigi =
+  let preprocess_intf sigi =
     match sigi with
     | { psig_desc = Psig_attribute attr; _ } :: rest
       when is_platform_tag attr.attr_name.txt ->
         if eval_attr attr = `keep then rest else []
     | _ -> List.filter_map apply_config_on_signature_item sigi
+
+  let traverse =
+    object (_ : Ast_traverse.map)
+      inherit Ast_traverse.map as super
+      method! structure expr = preprocess_impl (super#structure expr)
+      method! signature sigi = preprocess_intf (super#signature sigi)
+    end
 end
 
 let () =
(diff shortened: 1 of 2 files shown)


packages/browser-ppx/ppx.ml
```diff
@@ -510,19 +510,26 @@ module Preprocess = struct
         else None
     | Psig_extension _ | Psig_attribute _ -> Some sigi
 
-  let preprocess_impl _exp_ctxt str =
+  let preprocess_impl str =
```
davesnx, line 513: Super mega nit: I would inline both functions `preprocess_*`

davesnx (approved):
That's a great solution

packages/browser-ppx/ppx.ml
pedrobslisboa: Done: https://github.com/ml-in-barcelona/server-reason-react/pull/161/commits/fb404ceccc2c079d294180371427008d82c7bc84

Thank you 🎉

packages/browser-ppx/ppx.ml
```diff
@@ -510,19 +510,29 @@ module Preprocess = struct
...
+      inherit Ast_traverse.map as super
+
+      method! structure str =
+        let str = super#structure str in
+        match str with
+        | { pstr_desc = Pstr_attribute attr; _ } :: rest
+          when is_platform_tag attr.attr_name.txt ->
+            if eval_attr attr = `keep then rest else []
+        | str ->
+            List.filter_map apply_config_on_structure_item (super#structure str)
```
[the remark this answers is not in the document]
pedrobslisboa: You're right, It's a mistake that I didn't see.

davesnx (approved):

PR #168, opened 2024-08-22
Add native code to Js.String.split

pedrobslisboa:
## Description

This PR adds a native code to `Js.String.split`.
Before it was `notImplemented "Js.String" **"split"**`.

## How

`server-reason-react.js` has a melange behavior, so we need to keep it: 

#### API

Melange:
```ocaml
external split : ?sep:t -> ?limit:int -> t array = "split"
```

Native:
```ocaml
val split : ?sep:t -> ?limit:int -> t -> t array
```

### Interesting Cases 

I've separated some interesting cases regarding melange `String.split` behavior that was added to native code: 

- When `sep` matches some content at the beginning, it returns this match as an empty string  at the beginning of the list ([playground](https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiJvY2FtbCIgIm9jYW1sLnJlYXNvbi5yZWFjdCIp&live=off))
  Captura de Tela 2024-08-23 às 09 25 07

- When sep matches some content at the end, it returns this match as an empty string at the end of the list ([playground](https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiJyZWFjdCIgIm9jYW1sLnJlYXNvbi5yZWFjdCIp&live=off))
  Captura de Tela 2024-08-23 às 09 24 53

- When sep matches some content at the end and the beginning, it returns this match as an empty string at the end and beginning of the list ([playground](https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiIvIiAiL29jYW1sLnJlYXNvbi5yZWFjdC8iKQ%3D%3D&live=off))
  Captura de Tela 2024-08-23 às 09 31 23

- When sep is empty, it returns all the items ([playground](https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCAib2NhbWwucmVhc29uLnJlYWN0Iik%3D&live=off))
  Captura de Tela 2024-08-23 às 09 25 28

- When sep has no value, it returns all chars splitted ([playground](https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCAib2NhbWwucmVhc29uLnJlYWN0Iik%3D&live=off))
  Captura de Tela 2024-08-23 às 09 25 41

- Limit greater than array length returns all the items ([playground](https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiIuIiB%2BbGltaXQ6KDEwMCkgIm9jYW1sLnJlYXNvbi5yZWFjdCIp&live=off))
  Captura de Tela 2024-08-22 às 17 08 29

- Limit less than zero returns all the items ([playground](https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiIuIiB%2BbGltaXQ6KC0yKSAib2NhbWwucmVhc29uLnJlYWN0Iik%3D&live=off)) 
  Captura de Tela 2024-08-22 às 17 08 38

### DIFF
--- a/packages/melange.js/Js.ml
+++ b/packages/melange.js/Js.ml
@@ -758,28 +758,25 @@ end = struct
     else Stdlib.String.sub str start_idx (end_idx - start_idx)
 
   let split ?sep ?limit str =
-    match%platform () with
-    | Client -> Js.String.split ~sep ~limit str
-    | Server -> (
-        let sep = Option.value sep ~default:str in
-        let regexp = Str.regexp_string sep in
-        let items =
-          match
-            ( Str.string_match regexp str 0,
-              Str.string_match regexp str (String.length str - String.length sep),
-              sep <> "" )
-          with
-          | true, true, true -> ("" :: Str.split regexp str) @ [ "" ]
-          | true, false, true -> "" :: Str.split regexp str
-          | false, true, true -> Str.split regexp str @ [ "" ]
-          | _ -> Str.split regexp str
-        in
-        let items = items |> Stdlib.Array.of_list in
-        let limit = Option.value limit ~default:(Stdlib.Array.length items) in
-        match limit with
-        | limit when limit >= 0 && limit < Stdlib.Array.length items ->
-            Stdlib.Array.sub items 0 limit
-        | _ -> items)
+    let sep = Option.value sep ~default:str in
+    let regexp = Str.regexp_string sep in
+    let items =
+      match
+        ( Str.string_match regexp str 0,
+          Str.string_match regexp str (String.length str - String.length sep),
+          sep <> "" )
+      with
+      | true, true, true -> ("" :: Str.split regexp str) @ [ "" ]
+      | true, false, true -> "" :: Str.split regexp str
+      | false, true, true -> Str.split regexp str @ [ "" ]
+      | _ -> Str.split regexp str
+    in
+    let items = items |> Stdlib.Array.of_list in
+    let limit = Option.value limit ~default:(Stdlib.Array.length items) in
+    match limit with
+    | limit when limit >= 0 && limit < Stdlib.Array.length items ->
+        Stdlib.Array.sub items 0 limit
+    | _ -> items
 
   let splitByRe ~regexp ?limit str =
     let rev_array arr =
--- a/packages/melange.js/Js.ml
+++ b/packages/melange.js/Js.ml
@@ -760,15 +760,19 @@ end = struct
   let split ?sep ?limit str =
     let sep = Option.value sep ~default:str in
     let regexp = Str.regexp_string sep in
+    let str_start_with_sep = Stdlib.String.starts_with ~prefix:sep str in
+    let str_end_with_sep = Stdlib.String.ends_with ~suffix:sep str in
     let items =
-      match
-        ( Str.string_match regexp str 0,
-          Str.string_match regexp str (String.length str - String.length sep),
-          sep <> "" )
-      with
+      (* Melange (JS) has some specific cases for splitting strings: *)
+      (* https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiJzdGFydCIgInN0YXJ0Lm9jYW1sLnJlYXNvbi5yZWFjdCIpOwpKcy5sb2coSnMuU3RyaW5nLnNwbGl0IH5zZXA6ImVuZCIgIm9jYW1sLnJlYXNvbi5yZWFjdC5lbmQiKTsKSnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiJib3RoIiAiYm90aC5vY2FtbC5yZWFzb24ucmVhY3QuYm90aCIpOw%3D%3D&live=off *)
+      match (str_start_with_sep, str_end_with_sep, sep <> "") with
+      (* If the string starts and ends with the separator, we need to add an empty string at the beginning and the end *)
       | true, true, true -> ("" :: Str.split regexp str) @ [ "" ]
+      (* If the string starts with the separator, we need to add an empty string at the beginning *)
       | true, false, true -> "" :: Str.split regexp str
+      (* If the string ends with the separator, we need to add an empty string at the end *)
       | false, true, true -> Str.split regexp str @ [ "" ]
+      (* If the string doesn't start or end with the separator, we just split it *)
       | _ -> Str.split regexp str
     in
     let items = items |> Stdlib.Array.of_list in
--- a/packages/melange.js/Js.ml
+++ b/packages/melange.js/Js.ml
@@ -757,7 +757,29 @@ end = struct
     if start_idx >= end_idx then ""
     else Stdlib.String.sub str start_idx (end_idx - start_idx)
 
-  let split ?sep ?limit _str = notImplemented "Js.String" "split"
+  let split ?sep ?limit str =
+    match%platform () with
+    | Client -> Js.String.split ~sep ~limit str
+    | Server -> (
+        let sep = Option.value sep ~default:str in
+        let regexp = Str.regexp_string sep in
+        let items =
+          match
+            ( Str.string_match regexp str 0,
+              Str.string_match regexp str (String.length str - String.length sep),
+              sep <> "" )
+          with
+          | true, true, true -> ("" :: Str.split regexp str) @ [ "" ]
+          | true, false, true -> "" :: Str.split regexp str
+          | false, true, true -> Str.split regexp str @ [ "" ]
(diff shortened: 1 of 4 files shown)


packages/melange.js/Js.ml
```diff
@@ -757,7 +757,29 @@ end = struct
     if start_idx >= end_idx then ""
     else Stdlib.String.sub str start_idx (end_idx - start_idx)
 
-  let split ?sep ?limit _str = notImplemented "Js.String" "split"
+  let split ?sep ?limit str =
+    match%platform () with
+    | Client -> Js.String.split ~sep ~limit str
+    | Server -> (
```
davesnx, line 763: This lib is only native, so no need to use platform ppx here

packages/melange.js/Js.ml
```diff
@@ -757,7 +757,29 @@ end = struct
...
+        let sep = Option.value sep ~default:str in
+        let regexp = Str.regexp_string sep in
+        let items =
+          match
+            ( Str.string_match regexp str 0,
+              Str.string_match regexp str (String.length str - String.length sep),
+              sep <> "" )
+          with
+          | true, true, true -> ("" :: Str.split regexp str) @ [ "" ]
+          | true, false, true -> "" :: Str.split regexp str
```
davesnx, line 773: Could you add comments to those cases? They look either edge-cases or rarities of JS

packages/melange.js/dune
```diff
@@ -9,7 +9,7 @@
  (public_name server-reason-react.js)
  (libraries quickjs lwt str)
  (preprocess
-  (pps lwt_ppx)))
+  (pps lwt_ppx browser_ppx)))
```
davesnx, line 12: Yep, no need to add the browser_ppx!

packages/melange.js/test.ml
```diff
@@ -302,15 +302,44 @@ let string_tests =
         (* assert_string (Js.String.sliceToEnd ~from:(-2) "abcdefg") "fg"; *)
         assert_string (Js.String.slice ~start:7 "abcdefg") "");
     test "split" (fun () ->
```
davesnx, line 304: 🏅

packages/melange.js/Js.ml
pedrobslisboa: 😅 That make sense

pedrobslisboa: Done https://github.com/ml-in-barcelona/server-reason-react/pull/168/commits/cf71292de99b13fa9058fc72741a9608f11b1957

packages/melange.js/Js.ml
pedrobslisboa: Done: https://github.com/ml-in-barcelona/server-reason-react/pull/168/commits/b72485f7da48648848c435a87cdf79d1d185993f

packages/melange.js/Js.ml
```diff
@@ -757,7 +757,30 @@ end = struct
...
+  let split ?sep ?limit str =
+    let sep = Option.value sep ~default:str in
+    let regexp = Str.regexp_string sep in
+    let str_start_with_sep = Stdlib.String.starts_with ~prefix:sep str in
+    let str_end_with_sep = Stdlib.String.ends_with ~suffix:sep str in
+    let items =
+      (* Melange (JS) has some specific cases for splitting strings: *)
+      (* https://melange.re/unstable/playground/?language=OCaml&code=SnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiJzdGFydCIgInN0YXJ0Lm9jYW1sLnJlYXNvbi5yZWFjdCIpOwpKcy5sb2coSnMuU3RyaW5nLnNwbGl0IH5zZXA6ImVuZCIgIm9jYW1sLnJlYXNvbi5yZWFjdC5lbmQiKTsKSnMubG9nKEpzLlN0cmluZy5zcGxpdCB%2Bc2VwOiJib3RoIiAiYm90aC5vY2FtbC5yZWFzb24ucmVhY3QuYm90aCIpOw%3D%3D&live=off *)
+      match (str_start_with_sep, str_end_with_sep, sep <> "") with
+      (* If the string starts and ends with the separator, we need to add an empty string at the beginning and the end *)
```
davesnx, line 769: That makes sense, right? I wouldn't not say it's a edge case

```
splitting "." "1.2.3" -> ["1", "2", "3"]
splitting "1" "1.2.3" -> ["", ".2.3"]
```

When you split you always have an empty string if it matches at the start, since it's empty.

I wonder if the `str_start_with_sep` and `str_end_with_sep` are needed

pedrobslisboa: > When you split you always have an empty string if it matches at the start, since it's empty.

Not on ocaml [Str.split](https://ocaml.org/manual/5.2/api/Str.html#VALsplit)
> An occurrence of the delimiter at the beginning or at the end of the string is ignored.

With this code:
```ocaml
  let split ?sep ?limit str =
    let sep = Option.value sep ~default:str in
    let regexp = Str.regexp_string sep in
    let items = Str.split regexp str |> Stdlib.Array.of_list in
    let limit = Option.value limit ~default:(Stdlib.Array.length items) in
    match limit with
    | limit when limit >= 0 && limit < Stdlib.Array.length items ->
        Stdlib.Array.sub items 0 limit
    | _ -> items
```

Running this test:
https://github.com/ml-in-barcelona/server-reason-react/blob/11d6afce76b24ec8d8dd9f633a9d0a2717ff9897/packages/melange.js/test.ml#L317-L319

This is the output:
```
FAIL should be equal

   Expected: `[|""; "-sep-equals-to-beginning"|]'
   Received: `[|"-sep-equals-to-beginning"|]'
```

davesnx: Right, I see on the docs they mention: `An occurrence of the delimiter at the beginning or at the end of the string is ignored.` on Str.split: https://ocaml.org/manual/5.2/api/Str.html#VALsplit_delim:~:text=An%20occurrence%20of%20the%20delimiter%20at%20the%20beginning%20or%20at%20the%20end%20of%20the%20string%20is%20ignored.

But split_delim keeps the occurrences: https://ocaml.org/manual/5.2/api/Str.html#VALsplit_delim Do you think is worth changing?

pedrobslisboa: Thank you @davesnx, I didn't know `split_delim`. It totally worth add it instead of having this complex logic.

I just had to change one thing, with `split_delim` this happened:
```ocaml
  let split ?sep ?limit str =
    let sep = Option.value sep ~default:str in
    let regexp = Str.regexp_string sep in
    let items = Str.split_delim regexp str |> Stdlib.Array.of_list in
    let limit = Option.value limit ~default:(Stdlib.Array.length items) in
    match limit with
    | limit when limit >= 0 && limit < Stdlib.Array.length items ->
        Stdlib.Array.sub items 0 limit
    | _ -> items

(* test.ml *)
assert_string_array
  (Js.String.split ~sep:"" "with-empty-sep")
  [|
    "w"; "i"; "t"; "h"; "-"; "e"; "m"; "p"; "t"; "y"; "-"; "s"; "e"; "p";
  |];
```

```
FAIL should be equal

   Expected: `[|"w"; "i"; "t"; "h"; "-"; "e"; "m"; "p"; "t"; "y"; "-"; 
                "s"; "e"; "p"|]'

   Received: `[|"w"; "i"; "t"; "h"; "-"; "e"; "m"; "p"; "t"; "y"; "-"; 
                "s"; "e"; "p"; ""|]'
```

But with a empty `sep` in Js, [it doesn't add an empty string on the end](https://github.com/ml-in-barcelona/server-reason-react/pull/168#:~:text=When%20sep%20has%20no%20value%2C%20it%20returns%20all%20chars%20splitted%20(playground)). So I was thinking on do this:
```diff

  let split ?sep ?limit str =
    let sep = Option.value sep ~default:str in
    let regexp = Str.regexp_string sep in
-   let items = Str.split_delim regexp str |> Stdlib.Array.of_list in
+   let split = if sep <> "" then Str.split_delim else Str.split in
+   let items = split regexp str |> Stdlib.Array.of_list in
    let limit = Option.value limit ~default:(Stdlib.Array.length items) in
    match limit with
    | limit when limit >= 0 && limit < Stdlib.Array.length items ->
        Stdlib.Array.sub items 0 limit
    | _ -> items
```

WDYT?

davesnx: That make sense! Thanks for the back and forth. gj

PR #169, opened 2024-10-02
feat: add nested mel.obj support

pedrobslisboa:
## Description

Melange support nested `mel.obj`.
Example:
<table>
<tr>
<td> OCaml </td> <td> Reasonml </td>
</tr>
<tr>
<td>

```ocaml
let person = [%mel.obj { data = ([%mel.obj { name = "Pedro" }]) }]
```

</td>
<td>

```reason
let person = {
  "data": {
    "name": "Pedro",
  },
};
```

</td>
</tr>
</table>

We can see it working in the [Playground](https://melange.re/unstable/playground/?language=Reason&code=bGV0IHBlcnNvbiA9IHsKICAiZGF0YSI6IHsKICAgICJuYW1lIjogIlBlZHJvIiwKICB9LAp9OwoKSnMubG9nKHBlcnNvbiMjZGF0YSMjbmFtZSkK&live=off)

So this PR adds the same support to native `mel.obj`, as if you try to use it, it breaks:

Captura de Tela 2024-10-02 às 15 40 51

## How to test

- Access any `reasonml/ocaml` file in the demo
- Add a nested Js.t object, like the example above
- See that no error is shown

### DIFF
--- a/packages/melange.ppx/ppx.ml
+++ b/packages/melange.ppx/ppx.ml
@@ -765,6 +765,7 @@ class raise_exception_mapper (module_path : string) =
     inherit Ast_traverse.map as super
 
     method! expression expr =
+      let expr = super#expression expr in
       match expr.pexp_desc with
       | Pexp_extension
           ( { txt = "mel.obj"; _ },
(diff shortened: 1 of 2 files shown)


packages/melange.ppx/ppx.ml
```diff
@@ -765,6 +765,7 @@ class raise_exception_mapper (module_path : string) =
     inherit Ast_traverse.map as super
```
pedrobslisboa, line 765: This traverse is called `validate_record_labels`, but it works on `mel.*` extensions. This name seems weird to me because the context is applied.

davesnx:
Build on the demo is failing, but tests are passing. Merging and will fix macos failure on #main 

Thanks for this PR

PR #171, opened 2024-10-15
Handle DomProps with both jsx and html attribute name

pedrobslisboa:
## Description

Before, we had the HTML attribute for those `React.JSX.<type>` names, as the JSX isn't needed for SSR.
However, we must have this JSX attribute name for the RSC component.

## How

This PR adds to the DomProps the JSX attribute name data for `React.JSX.Bool` and `React.JSX.String`
Instead of holding only the HTML attribute name, `React.JSX.Bool` and `React.JSX.String` now hold both the HTML attribute name and the JSX attribute.

```diff
  type prop =
-    | Bool of (string * bool)
+    | Bool of ((string * string) * bool)
-    | String of (string * string)
+    | String of ((string * string) * string)
     | Style of string
     | DangerouslyInnerHtml of string
     | Ref of Ref.t
     | Event of string * event
```

### Discussion

@davesnx We talked about it to be a tuple, but could it be a record for a better read? What do you think?
```ocaml
  type prop_name = {
     html: string;
     jsx: string
  }

  type prop =
    | Bool of (prop_name * bool)
    | String of (prop_name * string)
    | Style of string
    | DangerouslyInnerHtml of string
    | Ref of Ref.t
    | Event of string * event
```

## How to test

- Run `make test-watch` into this branch
- Change whatever you want related to this subject, E.g.: 
  - Breaking the code: 
  ```diff
  -  React.JSX.String (("data-user-path", "data-user-path"), "what/the/path");
  +  React.JSX.String (("data-user-path", "data-user-path-error"), "what/the/path");
  ```
- Run `make demo` and see that nothing was changed


packages/react/src/DomProps.ml
```diff
@@ -1659,7 +1659,14 @@ let camelcaseToKebabcase str =
   in
   str |> chars_of_string |> loop [] |> List.rev |> string_of_chars
 
-let findByName tag jsxName =
+let findByName name =
```
davesnx, line 1662: Where this is being used?

packages/react/src/React.ml
```diff
@@ -339,8 +339,8 @@ module JSX = struct
     | Inline of string
 
   type prop =
-    | Bool of (string * bool)
-    | String of (string * string)
+    | Bool of ((string * string) * bool)
```
davesnx, line 342: I would use `string * string * bool` instead, wdyt?

packages/react/src/React.ml
```diff
@@ -339,8 +339,8 @@ module JSX = struct
...
+    | String of ((string * string) * string)
```
davesnx, line 343: Same here, not big reason to create a new tuple

packages/reactDom/src/ReactDOM.ml
```diff
@@ -249,7 +250,19 @@ let add kind value map =
 type dangerouslySetInnerHTML = < __html : string >
 
 (* `Booleanish_string` are JSX attributes represented as boolean values but rendered as strings on HTML https://github.com/facebook/react/blob/a17467e7e2cd8947c595d1834889b5d184459f12/packages/react-dom-bindings/src/server/ReactFizzConfigDOM.js#L1165-L1176 *)
-let booleanish_string name v = React.JSX.string name (string_of_bool v)
+let getPropByName name =
+  match DomProps.findByName name with
+  | Ok prop -> (name, DomProps.getJSXName prop)
+  | Error _ -> failwith "Invalid prop"
```
davesnx, line 256: can you move this into domprops, and remove the failwith?

packages/reactDom/src/ReactDOM.ml
```diff
@@ -249,7 +250,19 @@ let add kind value map =
...
+
+let string name v = React.JSX.string (getPropByName name) v
+let int name v = React.JSX.int (getPropByName name) v
+let bool name v = React.JSX.bool (getPropByName name) v
+
+let booleanish_string name v =
+  React.JSX.string (getPropByName name) (string_of_bool v)
+
+let float name v = React.JSX.float (getPropByName name) v
```
davesnx, line 265: We really don't want this, getPropByName is more expensive than the direct name, we always prefer to keep it static here

packages/react/src/React.ml
davesnx: It's better to keep a single tuple for perf, I haven't tested and just guessing but contextualisation here doesn't make much difference

packages/react/src/DomProps.ml
pedrobslisboa: It was used on `getPropByName` for `domProps` on `ReactDom`. But removed due to this comment: https://github.com/ml-in-barcelona/server-reason-react/pull/171#discussion_r1804281202

packages/reactDom/src/ReactDOM.ml
pedrobslisboa: It was removed due to this comment: https://github.com/ml-in-barcelona/server-reason-react/pull/171#discussion_r1804281202

PR #175, opened 2024-10-26
doc: add @@platform attribute documentation

pedrobslisboa:
## Description

Add documentation related to `@@platform` attribute

### DIFF
--- a/doc/browser_only.mld
+++ b/doc/browser_only.mld
@@ -63,6 +63,29 @@ switch%platform (Runtime.platform) {
 };
 ]}
 
+{2 @platform attribute}
+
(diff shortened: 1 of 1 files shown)


davesnx:
Good addition indeed, what do you think about having a ocaml version and a reason version:

```
[@platform js]
module X = {
  type t = Js.Json.t;

  let a = 2 + 2;
};
```

pedrobslisboa:
> [@platform js]
> module X = {
>   type t = Js.Json.t;
> 
>   let a = 2 + 2;
> };

@davesnx I changed it to be the only reason, as the entire doc is with reason. In the future, we can add both samples through the doc. WDYT?

davesnx:
Would be nice, but odoc doesn't support this transformation.

We might need to thing about another documentation website, any good ideas?

PR #176, opened 2024-10-29
Change setSearch to handle SearchParams.t

pedrobslisboa:
## Description

This PR changes the `setSearch` to handle SearchParams.t and add `setSearchString` to handle strings

## New API
```ocaml
let setSearchAsString: (t, string) => t;
let setSearch: (t, SearchParams.t) => t;
```


packages/url/native/URL.re
```diff
@@ -214,8 +214,12 @@ let search = url => {
   | _ => Some("?" ++ Uri.encoded_of_query(~scheme, query))
   };
 };
-let setSearch = (t, string) => {
-  Uri.with_query(t, Uri.query_of_encoded(string));
+let setSearchString = (t, searchString) => {
```
davesnx, line 217: I would rename this to `setSearchAsString`

davesnx (approved):
Tiny nit, but overall looks perfect

packages/url/native/URL.re
```diff
@@ -214,8 +214,12 @@ let search = url => {
...
+  Uri.with_query(t, Uri.query_of_encoded(searchString));
+};
+let setSearch = (t, searchParams) => {
+  let queryString = SearchParams.toString(searchParams);
+  setSearchString(t, queryString);
```
davesnx, line 222: no need to transform toString here, Uri.with_query should do the job, right?

pedrobslisboa: Good point

PR #177, opened 2024-10-30
Add useSyncExternalStore hooks

pedrobslisboa:
## Description 

The `useSyncExternalStore` and `useSyncExternalStoreWithServer` was missing on React hooks.

## How it works on server-reason-react 

For `useSyncExternalStore` it throws an error as react does not allowed to run it on a server without the `getServerSnapshot`. [Reference](https://github.com/facebook/react/blob/603e6108f39c6663ec703eed34a89ff1bf0cb70c/packages/react-server/src/ReactFizzHooks.js#L561-L566).

For `useSyncExternalStoreWithServer` we ignore all other params and call only the `getServerSnapshot` function.

Both follows the `react-reason` types: https://github.com/reasonml/reason-react/blob/c07e413750c5ca1632e6d3405a3cfbdf8d24c84e/src/React.rei#L192C1-L210C1

### DIFF
--- a/packages/react/src/React.ml
+++ b/packages/react/src/React.ml
@@ -570,6 +570,14 @@ let useState (make_initial_value : unit -> 'state) =
   in
   (initial_value, setState)
 
+type ('input, 'output) callback = 'input -> 'output
+
+let useSyncExternalStore ~subscribe:_ ~getSnapshot = getSnapshot ()
+
+let useSyncExternalStoreWithServer ~subscribe:_ ~getSnapshot:_
+    ~getServerSnapshot =
+  getServerSnapshot ()
+
 let internal_id = ref 0
 
 let useId () =
--- a/packages/react/src/React.mli
+++ b/packages/react/src/React.mli
@@ -632,6 +632,22 @@ val useCallback5 : 'a -> 'b -> 'a
 val useCallback6 : 'a -> 'b -> 'a
 val useId : unit -> string
 
+type ('input, 'output) callback = 'input -> 'output
+
+val useSyncExternalStore :
+  subscribe:((unit -> unit) -> (unit, unit) callback) ->
+  getSnapshot:(unit -> 'snapshot) ->
+  'snapshot
+[@@deprecated
+  "Use useSyncExternalStoreWithServer instead. More info at \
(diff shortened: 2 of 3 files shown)


packages/react/src/React.ml
```diff
@@ -570,6 +572,19 @@ let useState (make_initial_value : unit -> 'state) =
...
   (initial_value, setState)
 
+type ('input, 'output) callback = 'input -> 'output
+
+let useSyncExternalStore ~subscribe:_ ~getSnapshot:_ =
+  raise
+    (Not_implemented
+       "Do not use useSyncExternalStore on universal context, use \
+        useSyncExternalStoreWithServer instead. Read more about it \
+        https://github.com/facebook/react/blob/603e6108f39c6663ec703eed34a89ff1bf0cb70c/packages/react-server/src/ReactFizzHooks.js#L561-L566")
```
davesnx, line 582: What do you think about add a deprecation alert instead of raising an exception and return React.null?

davesnx: This way, there's no need to add Not_implemented

pedrobslisboa: @davesnx changed

pedrobslisboa:
@davesnx any other comment about it?

packages/react/src/React.mli
```diff
@@ -632,6 +632,22 @@ val useCallback5 : 'a -> 'b -> 'a
...
 val useId : unit -> string
 
+type ('input, 'output) callback = 'input -> 'output
+
+val useSyncExternalStore :
+  subscribe:((unit -> unit) -> (unit, unit) callback) ->
+  getSnapshot:(unit -> 'snapshot) ->
+  'snapshot
+[@@deprecated
+  "Use useSyncExternalStoreWithServer instead. More info at \
```
davesnx, line 642: I'm getting very picky, but just inline their comment no?

```
'Missing getServerSnapshot, which is required for server-rendered content.',
```

davesnx (approved):

packages/react/src/React.mli
pedrobslisboa: > I'm getting very picky
Not a problem to me 😁

PR #181, opened 2024-11-04
Update reason dependency

pedrobslisboa:
## Description

This PR updates the reason to 3.11.0 and higher and formats the files.
It attempts to solve the format issues on CI that required us to fix the reason to `3.10.0`.

@davesnx after I cleaned the opam cache and update to `3.11.0` I was able to run `make format` and see the output we have in this PR.

I have to say that one of the changes was strange to me, the change to raw:
```diff
- let%browser_only mockInitWebsocket = () => [%mel.raw
+ let%browser_only mockInitWebsocket = () =>
+  {%mel.raw |
    function mockInitWebsocket() {
      console.log("Load JS");
    }
- |}
- ];
+ |};
```
I don't know if we want to have or change this in some way.

## Why to have this PR?

In theory, this PR will let us continue without the reason dep blocked. 
And then let us, for example, to update `server-reason-react` on `styled-ppx`
Captura de Tela 2024-11-04 às 19 59 39

### DIFF
--- a/demo/universal/native/lib/MelRaw.re
+++ b/demo/universal/native/lib/MelRaw.re
@@ -1,13 +1,12 @@
-let%browser_only mockInitWebsocket = () => [%mel.raw
-  {|
+let%browser_only mockInitWebsocket = () =>
+  {%mel.raw |
   function mockInitWebsocket() {
     console.log("Load JS");
   }
-|}
-];
+|};
 
-let%browser_only initWebsocket = () => [%mel.raw
-  {|
+let%browser_only initWebsocket = () =>
+  {%mel.raw |
        function initWebsocket() {
          var socketUrl = "ws://" + location.host + "/_livereload";
          var s = new WebSocket(socketUrl);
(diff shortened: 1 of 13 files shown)


demo/universal/native/lib/MelRaw.re
```diff
@@ -1,13 +1,12 @@
...
     console.log("Load JS");
   }
-|}
-];
+|};
 
-let%browser_only initWebsocket = () => [%mel.raw
-  {|
+let%browser_only initWebsocket = () =>
+  {%mel.raw |
```
davesnx, line 9: Let's wait to merge this PR, until this is release: https://github.com/reasonml/reason/pull/2809

pedrobslisboa: Excellent, I found it weird as well.

pedrobslisboa: @davesnx Can we pin it until the release?

davesnx: Sure, there are no pins in this repo but go for it

pedrobslisboa: For anyone reading this, we decided to wait for the update.

PR #185, opened 2024-11-18
Add @browser_only

pedrobslisboa:
## Description

Working on removing platform code from expressions and pattern

## Why

A common problem when using `browser_only` is that we need to disable some warnings due to a present code on the native:

```ocaml
let%browser_only foo = () => "Bar"

(* JS result *)
let%browser_only foo = () => "Bar"

(* Native result *))
let%browser_only foo = () => 
      Runtime.fail_impossible_action_in_ssr "fun_value_binding_pexp_fun_2arg"
        [@@alert "-browser_only"]
```

So we still have `foo` on native. It only raises an error at runtime, and as `foo` should only be used on browsers, the compiler will complain of a non-used variable.

Another issue with browser_only is that it only works on functions, limiting its usage for other cases, such as args, variables, etc.

To work around it, I incremented `browser_only` to work as an attribute, which allowed us to work on other cases.
Example:
```reason
let foo = (~x as [@browser_only] x, y [@platform js]) => {
  let (a, [@browser_only b) = (42, 24);
  
  switch%platform () {
    | Client => a + b
    | Server => a
  }
}

// JS result
let foo = (~x as [@browser_only] x, y [@browser_only]) => {
  let (a, [@browser_only] b) = (42, 24);
  
  a + b
}

// Native result 
let foo = (~x as  _, _) => {
  let (a, _) = (42, 24);
  
  a
}
```

The ugliest part is that ocaml doesn't allow (At least I didn't find a way) to add an attribute directly to labeled args, so we need to do `(~x as [@browser_only] x` in reason and `~x:(x[@browser_only])` in ocaml 🫠. 

@davesnx @jchavarri Tagging, you guys, to discuss it. With this feature, I was able to remove the warning disables


packages/browser-ppx/tests/at_platform.t
```diff
@@ -124,6 +124,71 @@ Pstr_value
...
+  $ cat > input_let.ml << EOF
+  > let x =
+  >   let y = 42 [@@platform js] in
+  >   let y = 44 [@@platform native] in
+  >   y
+  > EOF
+
+  $ ./standalone.exe -impl input_let.ml | ocamlformat - --enable-outside-detected-project --impl
+  let x =
+    let y = 44 [@@platform native] in
```
davesnx, line 138: platform attribute should not be part of the output right?

packages/browser-ppx/tests/at_platform.t
```diff
@@ -124,6 +124,71 @@ Pstr_value
...
+  let x =
+    let y = 42 [@@platform js] in
+    y
+
+Pexp_let
+
+  $ cat > input_let.ml << EOF
+  > let x =
+  >   let y = 42 [@@platform js] in
+  >   let y = 44 [@@platform native] in
```
davesnx, line 151: This snapshot is indeed interesting because currently, the switch%platform makes more sense in those scenarios, right? 

I know that it's just a test, but trying to understand better how it works

packages/browser-ppx/tests/at_platform.t
```diff
@@ -124,6 +124,71 @@ Pstr_value
...
+
+  $ cat > input_let.ml << EOF
+  > let (x [@platform native], y [@platform js]) = (42, 44)
+  > EOF
+
+  $ ./standalone.exe -impl input_let.ml | ocamlformat - --enable-outside-detected-project --impl
+  let (x [@platform native]), _ = (42, 44)
+
+  $ ./standalone.exe -impl input_let.ml -js | ocamlformat - --enable-outside-detected-project --impl
+  let _, (y [@platform js]) = (42, 44)
```
davesnx, line 175: This case is indeed cool, but extremly rare in practice, right?

davesnx (commented):
I like the idea behind this PR, but it collides with the switch%platform that is a bit more explicit than the annotations.

Looking at the tests I see there's no clear use-case that this enables that isn't possible with a switch expression, so I would rather ask to add more test cases to expose clearly the patterns.

Otherwise looks great to add

packages/browser-ppx/tests/at_platform.t
pedrobslisboa: Yep, in case you want to handle a single variable as `y` the `switch%platform` is the way:
```reason
let y = switch%platform () {
  | Client => 42
  | Server => 44
}
```

But there will be cases which we just want to say that this code is "browser_only".
Example:
```reason
let make = (~initial, ~onClick as [@platform js] onClick) => {
  let (count, [@platform js] setCount) = RR.useStateValue(initial);

  [@platform js]
  let onClick = e => {
    setCount(count + 1);

    Js.log("This prints to the console");
    Js.log2("Printing count", count);

    onClick(e);
  };

  //...
```

In this case, for example, onClick is not required on serve, so we ca remove it from native code, `setCount` is also `browser_only` so we change it to `_` on native, same with `onClick`
I don't know if it's possible to have `switch%platform` on those cases, for example:
```reason
switch%platform () {
  | Client =>
    let onClick = e => {
      setCount(count + 1);

      Js.log("This prints to the console");
      Js.log2("Printing count", count);

      onClick(e);
    };
    ();
  | Server => ()
};

// Outputs the code bellow, which will not work:
{
  let onClick = e => {
    setCount(count + 1);
    Js.log("This prints to the console");
    Js.log2("Printing count", count);
    onClick(e);
  };

  ();
};
```

I think `%browser_only` would cover the function (will still have the warning) but not the tuple of even the prop.
The prop is a very common case. It is pretty normal to have onClick props on a CustomComponent, as shown in the code above.

packages/browser-ppx/tests/at_platform.t
pedrobslisboa: The platform attr is part of the output on all the examples of this test. Like this one:
```
Nested

  $ cat > input.ml << EOF
  >  module X = struct
  >    include struct
  >    type t = Js.Json.t
  >    let a = 2 + 2
  >    end [@@platform js]
  >  
  >    include struct
  >      type t = Js.Json.t
  >      let a = 4 + 4
  >    end [@@platform native]
  >  end
  > EOF

With -js flag it picks the block with `[@@platform js]`

  $ ./standalone.exe -impl input.ml -js | ocamlformat - --enable-outside-detected-project --impl
  module X = struct
    include struct
      type t = Js.Json.t
  
      let a = 2 + 2
    end [@@platform js]
  end

Without -js flag, it picks the block with `[@@platform native]`

  $ ./standalone.exe -impl input.ml | ocamlformat - --enable-outside-detected-project --impl
  module X = struct
    include struct
      type t = Js.Json.t
  
      let a = 4 + 4
    end [@@platform native]
  end
```

pedrobslisboa:
> I like the idea behind this PR, but it collides with the switch%platform that is a bit more explicit than the annotations.
> 
> Looking at the tests I see there's no clear use-case that this enables that isn't possible with a switch expression, so I would rather ask to add more test cases to expose clearly the patterns.
> 
> Otherwise looks great to add

I agree that the [@platform js] is not good for those cases; what about the `%browser_only` not being only an extension but also an attribute?
```reason
[@browser_only]
let make = () => {
  let (a, [@browser_only] y) = (42, 44)
}
```

pedrobslisboa:
@davesnx By the way, this PR is an attempt to improve what we already had with the `[@platform]`
For example this was already available:
```
Pstr_value

  $ cat > input_let.ml << EOF
  > let x = 42 [@@platform js]
  > let y = 44 [@@platform native]
  > EOF

  $ ./standalone.exe -impl input_let.ml | ocamlformat - --enable-outside-detected-project --impl
  let y = 44 [@@platform native]

  $ ./standalone.exe -impl input_let.ml -js | ocamlformat - --enable-outside-detected-project --impl
  let x = 42 [@@platform js]
```

But we could not do something like:
```
[@react.component]
let make = () => {

  [@platform js]
  let onClick = () => {}
}
```

demo/server/dune
```diff
@@ -2,7 +2,6 @@
  (name server)
  (enabled_if
   (= %{profile} "dev"))
- (flags :standard -w -26-27) ; browser_only removes code form the server, making this warning necessary
```
pedrobslisboa, line 5: :3 🎉

packages/runtime/Runtime.ml
```diff
@@ -9,6 +9,23 @@ let fail_impossible_action_in_ssr fn =
 
 Here's the raw callstack:
 
+%s
+|}
+      fn raw_callstack
+  in
+  raise (Impossible_in_ssr (Printf.sprintf {|'%s' shouldn't run on the server|} fn))
+
+let fail_impossible_action_in_ssr_anonymous fn =
```
jchavarri, line 18: How is this different than the other function? The only difference I can see is that this one prints the function with `Function:` while the other just shows it at the beginning of the msg (`'%s' sohld only run on the client...)

demo/universal/native/lib/Counter.re
```diff
@@ -1,10 +1,19 @@
-[@warning "-27"];
+let make = (~initial, ~onClick as [@browser_only] onClick) => {
+  let (count, [@browser_only] setCount) = RR.useStateValue(initial);
 
-let make = (~initial) => {
-  let (state, setCount) = RR.useStateValue(initial);
+  switch%platform (Runtime.platform) {
+  | Server => print_endline("This prints to the server terminal")
+  | Client => print_endline("This prints to the console")
+  };
```
jchavarri, line 7: Is this needed?

packages/browser-ppx/ppx.ml
```diff
@@ -5,8 +5,9 @@ type target = Native | Js
 
 let mode = ref Native
 let browser_ppx = "browser_ppx"
+let browser_only = "browser_only"
```
jchavarri, line 8: Can be used in a few places to replace string literal (e.g. line 131)

packages/browser-ppx/tests/at_platform.t
jchavarri: imo we should consolidate to either `platform js` or `browser_only`. Keeping both is a bit confusing to me. If we decide for one, we could deprecate the other in one version and remove it in the next.

jchavarri (commented):
Very cool!

davesnx (commented):
Right now `[@browser_only]` gets transformed into "do nothing" while [%browser_only] gets transformed into raising, do you think we could unify both? Or make `[@browser]` not fail, while _only fails? 

I'm happy to merge as is, and keep this idea around until we solidify the concepts, no rush on pushing it now :D

demo/universal/native/lib/Counter.re
pedrobslisboa: @jchavarri This piece of code was present before, but @davesnx removed it, and on the merge, I didn't notice it. Going to remove it.

pedrobslisboa:
@jchavarri @davesnx

I updated it with the current comments of this PR and what was discussed on the meeting. 
There is a point that I would like to add:
I could set `Pexp_fun` to be replaced by `()` as I'm doing with `Pexp_apply and Pexp_constant`:
```diff
       method! expression expr =
         let expr = super#expression expr in
         match expr.pexp_desc with
-        | Pexp_apply _ | Pexp_constant _ ->
+        | Pexp_apply _ | Pexp_constant _ | Pexp_fun ->
             let loc = expr.pexp_loc in
             if should_keep expr.pexp_attributes = `keep then expr else [%expr ()]
         | Pexp_let (_, [ { pvb_attributes = attrs; _ } ], body) -> if should_keep attrs = `keep then expr else body
-        | Pexp_fun _ ->
-            if should_keep expr.pexp_attributes = `keep then expr
-            else
-              let rec inner expr' =
-                match expr'.pexp_desc with
-                | Pexp_fun (label, def, _, expression) ->
-                    Builder.pexp_fun ~loc:expr'.pexp_loc label def (Builder.ppat_any ~loc:expr'.pexp_loc)
-                      (inner expression)
-                | _ ->
-                    let loc = expr'.pexp_loc in
-                    [%expr
-                      Runtime.fail_impossible_action_in_ssr
-                        [%e Builder.estring ~loc (Astlib.Pprintast.string_of_expression expr)]]
-              in
-              inner expr
        | _ -> expr
```

But it would provide an error as we don't have `[@browser_only]` on the JSX elements functions:
```
File "demo/universal/native/lib/Counter.re", line 24, characters 18-51:
24 |           onClick={[@browser_only] e => onClick(e)}>
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This expression should not be a unit literal, the expected type is
       React.Event.Mouse.t -> unit
```

I don't know if it's valid to remove the arg on `server-reason-react-ppx` like this:
```diff
  | Event { type_ = Mouse; jsxName }, false ->
      [%expr
        Some
          (React.JSX.Event
-             ([%e make_string ~loc jsxName], React.JSX.Mouse ([%e attribute_value] : React.Event.Mouse.t -> unit)))]
+             ([%e make_string ~loc jsxName], React.JSX.Mouse (fun _ -> () : React.Event.Mouse.t -> unit)))]
```
because it could be confusing for the user to always add [@browser_only] to every event callback function. But we already have the browser only on `useEffect`, which also triggers an error when using a function without `browser_only`.
Example of the errors:
| Without [@browser_only] | With [@browser_only] |
|--------|--------|
| Captura de Tela 2024-11-29 às 17 19 35 | Captura de Tela 2024-11-29 às 17 20 10 |
| Captura de Tela 2024-11-29 às 17 22 30 | Captura de Tela 2024-11-29 às 17 22 41 |

pedrobslisboa:
@davesnx @jchavarri Any extra comments about it?

davesnx:
Let me know if you need help with the conflicts

doc/browser_only.mld
```diff
@@ -26,46 +26,65 @@ You would need to add it on both "server" and "client" dune files. Adding the [-
 ]}
 
 {2 browser_only}
+
+A code with [[@browser_only]] attribute will keep the function untouched for the client build, but for the server build, will be transformed into to a dead code. If you try to run it on the server, it will fail as it does not exist on the server, and the compiler will crash.
```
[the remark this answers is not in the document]
pedrobslisboa: But do you think an entire page is necessary? Can it be just a section in this one with a reference to %browser_only?

pedrobslisboa:
No need for conflict help.
I think I got everything work. Let me tell if you have any doubt

BTW I followed the way we handled "under the hoods" the use effect and did the same with React.Event handlers:
```diff
  | Event { type_ = Mouse; jsxName }, false ->
      [%expr
        Some
          (React.JSX.Event
-             ([%e make_string ~loc jsxName], React.JSX.Mouse ([%e attribute_value] : React.Event.Mouse.t -> unit)))]
+             ([%e make_string ~loc jsxName], React.JSX.Mouse (fun _ -> () : React.Event.Mouse.t -> unit)))]
```

pedrobslisboa:
Ok @davesnx and @jchavarri I was testing it on a real project to check for issue, and I found that replacing browser_only with () would lead to type issues, like:
Captura de Tela 2025-02-02 às 08 11 00
As you can see, on that if/else it has the type error due to unit and int incoherence

To workaround it, I'm working now with `Obj.magic` and `Obj.t` and also added compilation on cram tests to ensure our code is compiling well.

pedrobslisboa:
This is just another sample that we have the power of the warnings :/ With @browsert_only, we can get rid off it
Captura de Tela 2025-02-02 às 16 53 58

Issue #191, opened 2024-12-04
JSX syntax context not working well with context + provider value


pedrobslisboa:
## Description

Due to an issue with Context and initial value on Provider, I tried to understand why it was happening on the project (With JSX) but not on this repo.

To do it, I create a branch with a `test_context.re` file.
It's a simple test to check handle context with JSX syntax:
```ocaml
let context = React.createContext(10);

module ContextProvider = {
  include React.Context;

  let make = React.Context.provider(context);
};

module ContextConsumer = {
  [@react.component]
  let make = () => {
    let value = React.useContext(context);
    <section> {React.int(value)} </section>;
  };
};

let context = () => {
  let component =
    <ContextProvider value=20> <ContextConsumer /> </ContextProvider>;

  assert_string(
    ReactDOM.renderToStaticMarkup(component),
    "<section>20</section>",
  );
};
```
With this file, we can see the error:

```
ASSERT should be equal
FAIL should be equal

   Expected: `"<section>20</section>"'
   Received: `"<section>10</section>"'
```

When running the test like this:
```reason
let context = () => {
  let _component =
    <ContextProvider value=20> <ContextConsumer /> </ContextProvider>;

  let component =
    <ContextProvider value=30> <ContextConsumer /> </ContextProvider>;

  assert_string(
    ReactDOM.renderToStaticMarkup(component),
    "<section>20</section>",
  );
};
```

The test passes, showing that we only set the value on a second call, using the previous value, as the second on is `30`

I'm opening this issue while I'm working on this on the same branch I created: https://github.com/ml-in-barcelona/server-reason-react/tree/feat/context

pedrobslisboa:
Debugging, I found some things:

We have this code on testing:
```reason
module ContextProvider = {
  include React.Context;

  let make = React.Context.provider(context);
};

module ContextConsumer = {
  [@react.component]
  let make = () => {
    let value = React.useContext(context);
    <section> {React.int(value)} </section>;
  };
};

let context = () => {
  let component =
    <ContextProvider value=20> <ContextConsumer /> </ContextProvider>;

  assert_string(
    ReactDOM.renderToStaticMarkup(component),
    "<section>20</section>",
  );
};
```
This is the code after ppx:

```ocaml
module ContextConsumer = {
    let make ?key:(_ : string option)  () =
      let value = React.useContext context in
      React.Upper_case_component
        (fun () -> React.createElement "section" [] [React.int value])
};

let context () =
  let component =
[...]

PR #192, opened 2024-12-05
Move ppx transformation to function body (not last expression)

davesnx:
This PR fixes https://github.com/ml-in-barcelona/server-reason-react/issues/191

It moves the wrapping of `React.upper_case_ and React.async_` as a first expression on the make binding. Before this was added as the last expression found (if it was a sequence). 

The bug appears when `useContext` is used, since the execution of this hook would be in a different execution order from the rest of the component.


pedrobslisboa (approved):
Greate work, @davesnx, thank you. 🎉

PR #196, opened 2024-12-19
RSC syntax ppx

davesnx:
- Add `[@client]`
- Add encoding based on type annotations of props
- Fix demo using the new ppx
- Use `import(...).then(module => ({ default: module.make })))` for the demo and remove the `let default` on each client component


pedrobslisboa (commented):
Missed this review, but it looks perfect. Congrats 🎉

PR #199, opened 2025-01-17
Run extract-client-components as an esbuild plugin

davesnx:


packages/extract-client-components/extract_client_components.ml
```diff
@@ -88,33 +75,25 @@ let capture_all_client_component_files_in_target path =
...
+  Cmdliner.Arg.(required & pos 0 (some string) None & info [] ~docv:"MELANGE_TARGET" ~doc)
 
-let extract_components target path =
+let extract_components target =
   let current_dir = Sys.getcwd () in
   let melange_target = Filename.concat current_dir target in
   match capture_all_client_component_files_in_target melange_target with
   | Ok manifest ->
-      render_manifest ~path manifest;
+      print_endline (render_manifest manifest);
```
pedrobslisboa, line 87: Why do we need to print it?

pedrobslisboa (approved):
The more I see it, the more I get excited to see the final demo. Great work man ❤️ 🎉

packages/extract-client-components/extract_client_components.ml
davesnx: we print to the stdout so I can use the CLI from both:

- dune `(with-stdout-to (cli))`
- esbuild `const content = execSync("cli"); Fs.writeFile(...)`

Before the cli was writing the file, so from the outside had less control. That could be configured by a parameter, but the stdout should probably be the default

PR #215, opened 2025-02-27
Add hydrate information on doc

pedrobslisboa:


documentation/ssr-and-hydration.mld
```diff
@@ -0,0 +1,291 @@
...
+Commonly, hydrate errors will appear as:
+{%html:
+  <p style="background-color: #362422; color: #df5452; padding: 1em;">
+    <strong>Text content does not match server-rendered HTML.</strong>
+  </p>
+%}
+
+This means that while rendering the application, there was a difference between the React tree pre-rendered from the server and the React tree rendered during the first render in the browser (hydration).
+
+As it's hard to see it just in words, let's go to an example with this simple code:
```
davesnx, line 27: I would add a heading like:

"Example"
Let's see how we could trigger on purpose a hydration error, take a look at this code:

documentation/ssr-and-hydration.mld
```diff
@@ -0,0 +1,291 @@
...
+
+{%html:<img src="./ssr-hydrate-issue.png" alt='Hydrate error' style="max-width: 100%; height: auto; vertical-align: middle;">%}
+
+* As we can see, there is a difference between III and IV, that will cause React to complain about content mismatch.
+
+This is because React understands that the markup from the server ([background-color: red](IV)) differs from the one it expects to be on the client at the hydration time ([background-color: blue](V)). 
+
+If we want different values on the server and client, we must do it after React hydrates the client or, commonly said, after *mount*.
+
+To work on it we must check if the client mounted and then change this value. This can be done with a hook that will be executed only after the client is hydrated.
```
davesnx, line 52: Add a heading like, "Possible solution: mounting on the client only"

davesnx: and mention that ideally those hydration errors should be fixed, either by removing the difference or by adding this extra solutions

documentation/ssr-and-hydration.mld
```diff
@@ -0,0 +1,291 @@
...
+    None;
+  });
+
+  isMounted;
+};
+]}
+
+{3 When using it:}
+
+You must use [IsClientMounted] whenever you have to deal with variables inside a Component body (Not the markup).
```
davesnx, line 107: I don't understand this sentence ^^

pedrobslisboa: The idea was to say that `IsClientMounted` is for variables, not for React Markup, that would be handle by `ServerOrClientRender`. But I agree that it's confusing, I updated to:

```
You should use [IsClientMounted] whenever you have to deal with variables inside a component. 
For React markup, take a look at {{!section-"server-or-client-render"} <ServerOrClientRender/>} 
or {{!section-"client-only"} <ClientOnly/>}.

Example:

{@reasonml[
let isFocusable =
  isClientMounted ? children->isFocusableElement : false;
]}
```
image

documentation/ssr-and-hydration.mld
```diff
@@ -0,0 +1,294 @@
+{0 SSR and hydration}
+
+{1 How does React handle SSR markup?}
+During server-side rendering (SSR), React generates static HTML markup on the server. When this markup reaches the client's browser, React performs a process called "hydration", 
+it attaches event handlers and other interactive behaviors to the existing HTML elements, rather than re-rendering the entire page from scratc
+
+For hydration to work correctly, the initial server markup must match exactly what React expects on the client - otherwise React will be forced to re-render the page, negatively impacting performance.
```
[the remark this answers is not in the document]
pedrobslisboa: > Maybe worth explaining that hydration attaches the events the same way as react re-renders when a state changes, doing the vdom diff. (not sure what are the right words to explain this simply)

I don't know if adding this level of complexity to this documentation is worth it.

documentation/ssr-and-hydration.mld
```diff
@@ -0,0 +1,301 @@
...
+it attaches event handlers and other interactive behaviors to the existing HTML elements, rather than re-rendering the entire page from scratc
+
+For hydration to work correctly, the initial server markup must match exactly what React expects on the client. 
+If there is a mismatch, React will throw a hydration error.
+
+For example:
+- server renders:[<span>Hello from server</span>]
+- client renders:[<span>Hello from client</span>]
+
+With this mismatch, instead of being able to efficiently "hydrate" the existing server markup with interactivity, React will throw away the server-rendered content and do a full client-side render, resulting in worse performance and user experience, like layout shifts and other annoyances.
```
raphael-proust, line 14:
```suggestion
With this mismatch, instead of being able to efficiently hydrate the existing server markup with interactivity, React will throw away the server-rendered content and do a full client-side render, resulting in worse performance and user experience, like layout shifts and other annoyances.
```

You've used hydration/hydrate before, it's a defined technical jargon term now, you should drop the quotes.

documentation/ssr-and-hydration.mld
```diff
@@ -0,0 +1,301 @@
...
+
+** As we can see, there is no difference between III and IV, and React will be happy with it.
+
+In this case, we don't even need [switch%platform] to know if it's on client or native as [useEffects] only runs on client and the value is provided on the client and the content "blue" and "red" are native compatible.
+
+{1 Helpers that you can use}
+
+{2 IsClientMounted}
+
+IsClientMounted is a hook helper that you can create to check if the client is mounted.
```
raphael-proust, line 98:
```suggestion
[IsClientMounted] is a hook helper that you can create to check if the client is mounted.
```

for code

documentation/ssr-and-hydration.mld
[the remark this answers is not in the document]
davesnx: Make sense, but mention that isn't a "special react thing" and rather a common operation

PR #216, opened 2025-03-04
Fix comments for perf in Html.ml and React.ml

davesnx:
Thanks for the suggestions @raphael-proust


raphael-proust (approved):
nice work :+1:

PR #218, opened 2025-03-07
Create server_actions ppx

pedrobslisboa:
## Description

This PR is the first step of creating the server actions feature for SRR.
In this part, we focus on server actions imported inside `react.client.components`.
Also is focused on JSON response content to later support server actions for FormData and other forms of response data.

It was chosen to use attribute instead of extension because we need to add extra content in this `structure_item,` and extension isn't the feature for it.

## How

The ppx is used on `structure_item` functions, so a function like this:

```reason
[@react.server.action]
let some_action = (~a: string, ~b: int): Js.Promise.t(string) => {
  Promise.resolve(a ++ string_of_int(b));
};
```

Will result in:

```reason
// Client
let some_action = (~a: string, ~b: int) => {
  let location = Webapi.Dom.window |> Webapi.Dom.Window.location;
  let encodeArgs =
    "{"
    ++ String.concat(
          ",",
          List.map(
            ((key, value)) => key ++ ":" ++ value,
            [
              (
                "a",
                Ppx_deriving_json_runtime.to_string([%to_json: string](a)),
              ),
              ("b", Ppx_deriving_json_runtime.to_string([%to_json: int](b))),
            ],
          ),
        )
    ++ "}";
  let body = Fetch.BodyInit.make(encodeArgs);
  let basePath =
    switch (SRRServer.baseRoute) {
    | Some(baseRoute) => baseRoute
    | None => ""
    };
  Fetch.fetchWithInit(
    Webapi.Dom.Location.origin(location) ++ basePath ++ "623496469",
    Fetch.RequestInit.make(~method_=Post, ~credentials=Include, ~body, ()),
  )
  |> Js.Promise.then_(result =>
        try(Fetch.Response.json(result)) {
        | exn => Js.Promise.reject(exn)
        }
      )
  |> Js.Promise.then_(json => Js.Promise.resolve([%of_json: string](json)));
};

// Server
[@react.server.action]
let some_action = (~a: string, ~b: int): Js.Promise.t(string) =>
  Promise.resolve(a ++ string_of_int(b));

SRRServer.register_action(
  ~route={
    let basePath =
      switch (SRRServer.baseRoute) {
      | Some(baseRoute) => baseRoute
      | None => ""
      };
    basePath ++ "623496469";
  },
  ~handler=body => {
    let args = Yojson.Basic.from_string(body);
    let%lwt result =
      some_action(
        ~a=[%of_json: string](Yojson.Basic.Util.member("a", args)),
        ~b=[%of_json: int](Yojson.Basic.Util.member("b", args)),
      );
    Js.Promise.resolve(
      Ppx_deriving_json_runtime.to_string([%to_json: string](result)),
    );
  },
);
```
As we can see, we don't opine on any server tool, but the ppx requires the creation of an adapter called SRRServer to connect your server tool to the actions handler, for example:
```reason
// Defines a custom base route for the actions
// Isn't required but allows the user to define a specific path
let baseRoute = Some("/api/actions/");

// Defines the functions to adapt the server actions to the server
[@platform native]
include {
          let actionRoutes = ref([]);

          let register_action = (~handler, ~route) => {
            actionRoutes :=
              [
                Dream.post(
                  route,
                  req => {
                    let%lwt body = Dream.body(req);
                    let%lwt json_string = handler(body);
                    Dream.json(json_string);
                  },
                ),
                ...actionRoutes^,
              ];
          };

          let get_action_routes = () => actionRoutes^;
        };
```

Don't know if it's the best option.
@davesnx about this adapter, any idea?

### DIFF
--- a/demo/server/server.re
+++ b/demo/server/server.re
@@ -27,6 +27,7 @@ let server =
         Pages.SinglePageRSC.handler,
       ),
       Dream.get(Router.demoRouter, Pages.RouterRSC.handler),
+      ...SRRServer.get_action_routes(),
     ]),
   );
 
--- a/demo/universal/native/lib/Counter.re
+++ b/demo/universal/native/lib/Counter.re
@@ -1,10 +1,22 @@
 open Ppx_deriving_json_runtime.Primitives;
 
+[@react.server.action]
+let onClickServerAction =
+    (~foo as a: string, ~bar: int): Js.Promise.t(string) => {
+  Js.Promise.resolve(a ++ string_of_int(bar));
+};
+
 [@react.client.component]
 let make = (~initial: int) => {
   let (state, setCount) = RR.useStateValue(initial);
 
   let onClick = _event => {
+    onClickServerAction(~foo="foo", ~bar=1)
+    |> Js.Promise.then_(result => {
+         Js.log("############" ++ result);
+         Js.Promise.resolve();
+       })
+    |> ignore;
     setCount(state + 1);
   };
 
--- a/demo/universal/native/lib/SRRServer.re
+++ b/demo/universal/native/lib/SRRServer.re
@@ -1,0 +1,25 @@
+// Defines a custom base route for the actions
+let baseRoute = Some("/api/actions/");
+
+// Defines the functions to adapt the server actions to the server
+[@platform native]
+include {
+          let actionRoutes = ref([]);
+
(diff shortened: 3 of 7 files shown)


pedrobslisboa:
Starting from the beginning.

pedrobslisboa:
Archive, we check it later if any info from here is needed

PR #219, opened 2025-03-11
react-server-dom-esbuild-client

davesnx:
- Brings a new react-server-dom-esbuild-client module which uses react-client flight
- Current status of react-server-dom-esbuild-client is wip, but it currently works
- Updates react to 19.1
- Updates the ReactServerDOM RSC payload to support new expected format (I mock it to move this PR forward, I might push the good impl soon)
- Make tests faster by making the sleeps in ms not s


demo/universal/js/ReactServerDOMWebpack.re
pedrobslisboa, line 1: Do we need to keep this file? Can't we delete it as we are using the Esbuild module?

davesnx: I want to keep in case for a fast check

pedrobslisboa (approved):
Greate work!

PR #223, opened 2025-03-19
Improve old client_props

pedrobslisboa:
## Description

The idea of this PR is to improve the old client_props and start the movement to align the behavior of SRR RSC to what React and Next do.

## Why

The react expect that response of the RSC, which evolves actions response and props response to the client, to be composable.
```reason
let props = {
    bar: {
     element: <div/>
  },
  foo: "Hey",
  baz: [<div/>, "Yah!"]
}
```

But our type nowadays is:
```ocaml
and client_prop =
  (* TODO: Do we need to add more types here? *)
  | Json : Yojson.Basic.t -> client_prop
  | Element : element -> client_prop
  | Promise : 'a Js.Promise.t * ('a -> Yojson.Basic.t) -> client_prop
```

As we can see, `Json` does not support any other `client_prop` inside it, similar to `Promise` that does not support it. They always handle a `Yojson.Basic.t`.

So we cannot have: React.Promise(promise, (response) => React.Element(response)).
How would we have such response:
```reason
// Server content that needs to be serialized
let props = {
    bar: {
     element: <div> {React.String("I'm a nested element :3")} </div>
  },
  foo: [<div> {React.String("Hey")} </div>, "Yah!"],
  baz: <div> {React.String("Who is Lola?")} </div>
}

//Let's see it as client prop
{
  "bar":  React.???(["element", React.Element(<div> {React.String("I'm a nested element :3")} </div> )])
  "foo":  React.???([React.Element(<div> {React.String("Hey")} </div>, React.Json(`String("Yah!")))])
  "baz": React.Promise(React.Element(<div> {React.String("Who is Lola?")} </div>)) // This is not valid nowaday
)
```

We need to support those React.??? and we need to make Promise to accept `client_prop` instead only Yojson type.
My propose is to provide a more complete `client_prop`:
```ocaml
and client_prop =
  | List : client_prop list -> rsc_value
  | Assoc : (string * client_prop) list -> rsc_value
  | Json : Yojson.Basic.t -> client_prop
  | Element : element -> client_prop
  | Promise : 'a Js.Promise.t * ('a -> client_prop) -> client_prop
```

Then we can handle all those types recursively, and the final return must always be either a `Yojson.Basic.t` or a `React.element`.

```reason
{
  "bar":  React.Assoc(["element", React.Element(<div> {React.String("I'm a nested element :3")} </div> )])
  "foo":  React.List([React.Element(<div> {React.String("Hey")} </div>, React.Json(`String("Yah!")))])
  "baz": React.Promise(React.Element(<div> {React.String("Who is Lola?")} </div>))
)
```

As this PR looks to adapt it for both props response and server actions, I also think that client_props is not a valid name anymore, but `rsc_value`. But as it can cause some confusion, I'm moving it to the server-actions PR.

### The code contains some `question` comments that are also already pointed into my notion of the subsequent improvements.

### DIFF
--- a/packages/react/src/React.ml
+++ b/packages/react/src/React.ml
@@ -390,9 +390,11 @@ and client_props = (string * client_prop) list
 
 and client_prop =
   (* TODO: Do we need to add more types here? *)
+  | PropList : client_prop list -> client_prop
+  | Assoc : (string * client_prop) list -> client_prop
   | Json : Yojson.Basic.t -> client_prop
   | Element : element -> client_prop
-  | Promise : 'a Js.Promise.t * ('a -> Yojson.Basic.t) -> client_prop
+  | Promise : 'a Js.Promise.t * ('a -> client_prop) -> client_prop
 
 exception Invalid_children of string
 
--- a/packages/reactDom/src/ReactServerDOM.ml
+++ b/packages/reactDom/src/ReactServerDOM.ml
@@ -116,7 +116,7 @@ module Model = struct
       | Upper_case_component component ->
           let element = component () in
           (* Instead of returning the payload directly, we push it, and return a reference to it.
-             This is how `react-server-dom-webpack/server` renderToPipeableStream works *)
+           This is how `react-server-dom-webpack/server` renderToPipeableStream works *)
           let index = use_chunk_id context in
           context.push index (Chunk_value (to_payload element));
           `String (ref_value index)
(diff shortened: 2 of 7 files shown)


packages/react/src/React.ml
```diff
@@ -386,13 +386,15 @@ type element =
...
+and rsc_value =
   (* TODO: Do we need to add more types here? *)
-  | Json : Yojson.Basic.t -> client_prop
-  | Element : element -> client_prop
-  | Promise : 'a Js.Promise.t * ('a -> Yojson.Basic.t) -> client_prop
+  | RSC_value_List : rsc_value list -> rsc_value
+  | RSC_value_Assoc : (string * rsc_value) list -> rsc_value
+  | RSC_value_Json : Yojson.Basic.t -> rsc_value
+  | RSC_value_Element : element -> rsc_value
+  | RSC_value_Promise : 'a Js.Promise.t * ('a -> rsc_value) -> rsc_value
```
davesnx, line 397: We might need to change the name from client_prop to `value` or similar. 

What I'm not so sure is the idea to expose List/Assoc in here. Since Json is all we need.

`RSC_value_List xx` is the same as `RSC_value_Json 'List ()`

Also, a worth note about naming and casing. The prefix of `rsc_` is not ideal, and RSC_value into all variants is criminal.

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -96,92 +96,103 @@ module Model = struct
     Buffer.add_string buf "\n";
     Buffer.contents buf
 
+  let rec element_to_payload ~context element =
```
davesnx, line 99: I don't understand this abstraction, can you give some details?

packages/react/src/React.ml
pedrobslisboa: > What I'm not so sure is the idea to expose List/Assoc in here. Since Json is all we need.

We cannot use the Json type to compose with other types as Promise, for example.

This is not allowed:
```reasonml
let prop = {
  foo: "Hey",
  bar: {
     element: <div/>
  }
}
```

pedrobslisboa: > Also, a worth note about naming and casing. The prefix of rsc_ is not ideal, and RSC_value in all variants is criminal.

I agree. I just pushed with this name because I had no idea, and making and discussing it in this PR would have been better.

The value that I created was:
```
  | List : rsc_value list -> rsc_value
  | Assoc : (string * rsc_value) list -> rsc_value
  | Json : Yojson.Basic.t -> rsc_value
  | Element : element -> rsc_value
  | Promise : 'a Js.Promise.t * ('a -> rsc_value) -> rsc_value
```
But I got the error for the List cause `React.List` already exists, so I pushed this prefix to discuss here.

davesnx: > > What I'm not so sure is the idea to expose List/Assoc in here. Since Json is all we need.
> 
> We cannot use the Json type to compose with other types as Promise, for example.
> 
> This is not allowed:
> 
> ```
> let prop = {
>   foo: "Hey",
>   bar: {
>      element: <div/>
>   }
> }
> ```

Adding `Assoc` or similar would be sufficient to support it, right? No need to add List

pedrobslisboa: But then, how would we represent this:
```
let prop = {
  foo: "Hey",
  bar: [<div/>, "Yah!"]
}
```
?
Assoc is: (string, rsc_value) list
List is: rsc_value list

davesnx:
```
let props = {
  foo: "Hey",
  bar: [<div/>, "Yah!"]
}

let props = Assoc([("foo", 'String "Hey"), ("bar", 'List(...))])
```

Inside Assoc there're all the Json constructors. That's why I would design a type without even Assoc (and use Json + Promise + Element)

PR #224, opened 2025-03-19
Server action with demo

pedrobslisboa:
## Description

Create the minimum for a demo without new ppxs and minimum magic work.

This will count with the ability to:

- Declare actions
    - For the server: router and action_id
    - For the client: request function
- Pass an action on a prop/attribute
    - On the action attribute on a <form>
    - On a prop for the client component
        
- Execute those actions on the client with react tools

Todos

- [x]  Handle actions that are imported on client components
- [x]  Adapt ReactServerDOM to be able to handle functions.
- [x]  Handle FormData on action prop
    - [x]  Change server-reason-react.ppx, and DomProps to handle it.
- [x]  Handle actions on server component props
- [x]  Handle action on client component props
    - [x]  Change server-reason-react.ppx and React to handle it.


demo/server/DB.re
```diff
@@ -71,6 +71,63 @@ let findOne = (notes, id) => {
...
+          content,
+          updated_at: Unix.time(),
+        };
+        [note, ...notes];
+      },
+      notes,
+    );
+  Cache.set(notes);
+  Lwt_result.lift(notes |> Result.map(notes => notes |> List.hd));
+};
```
davesnx, line 92: This doesn't addNote into the DB, but rather into the Cache. Which is perfectly fine, but no need to do `Lwt_result.lift(notes |> Result.map(notes => notes |> List.hd));` right?

demo/universal/js/FetchHelpers.re
```diff
@@ -0,0 +1,65 @@
...
+                ReactServerDOMWebpack.createFromReadableStream(body);
+              })
+           |> ignore;
+         },
+         el |> Webapi.Dom.Element.asEventTarget,
+       )
+     });
+
+  ();
+};
```
davesnx, line 65: is this how is implemented to support progressive enhancement?

davesnx (commented):
I don't get why we have 2 PRs depending on each other. Reviewd a few nits but a few questions as well

packages/server-reason-react-ppx/DomProps.ml
```diff
@@ -688,7 +688,7 @@ let formHTMLAttributes =
     Attribute { name = "rel"; jsxName = "rel"; reasonJsxName = "rel"; type_ = String };
     Attribute { name = "enctype"; jsxName = "encType"; reasonJsxName = "encType"; type_ = String };
     Attribute { name = "action"; jsxName = "action"; reasonJsxName = "action"; type_ = String };
-    Attribute { name = "method"; jsxName = "method"; reasonJsxName = "method_"; type_ = String };
+    Attribute { name = "method"; jsxName = "method"; reasonJsxName = "method"; type_ = String };
```
pedrobslisboa, line 691: Still need to know how this is going to work on the SRR. But `reason-react` has it as "method"

pedrobslisboa: It's working on sample, but breaking on cram test

demo/universal/js/ReactServerDOMWebpack.re
```diff
@@ -0,0 +1,64 @@
+/* TODO: Move this bindings into reason-react */
+type callServerCallback('a, 'b) = (string, list('a)) => Js.Promise.t('b);
```
davesnx, line 2: could you rename 'a and 'b to be something meaningful? I applies to the entire module, I t takes a bit to follow all the a/b and createServerReferenceImpl generics

demo/universal/native/lib/Actions.re
```diff
@@ -0,0 +1,296 @@
...
+  Yojson.Basic.from_string(body)
+  |> Yojson.Basic.Util.to_list
+  |> List.hd
+  |> Yojson.Basic.Util.to_list;
+};
+
+module Notes = {
+  [@platform native]
+  let createNoteResponse = note => {
+    React.Json(
```
davesnx, line 16: I think React.Json shouldn't be used and should be wrapped by ppx, do you agree?

demo/universal/native/lib/Actions.re
```diff
@@ -0,0 +1,296 @@
...
+  let createRouteHandler = body => {
+    // Parse the body to get the args
+    let (title, content) =
+      switch (getArgs(body)) {
+      | [title, content] => (
+          // It would be handle by a title_of_json/content_of_json in the future provided by the end-user and the ppx
+          title |> Yojson.Basic.Util.to_string,
+          content |> Yojson.Basic.Util.to_string,
+        )
+      | _ => failwith(Printf.sprintf("Invalid arguments %s", body))
```
davesnx, line 53: Can you deal with errors in all the functions in case of wrong types at runtime?

demo/server/server_actions/Route.re
```diff
@@ -0,0 +1,81 @@
+// QUESTION: How should we create this manifest automatically?
```
davesnx, line 1: This is something we need to know before merging this PR, do you have a cool answer?

demo/server/server_actions/Route.re
```diff
@@ -0,0 +1,81 @@
...
+    switch (formData, actionId) {
+    // react-server-dom-webpack encode formData and put the first value as model reference E.g.:["$K1"], 1 is the id
+    // QUESTION: Why is it like this? Is there a change to it be other than "$K"? Is there a way to get more ids?
+    | ([(_, [(_, modelId)]), ...formData], Some(actionId)) =>
+      let chunkId = Yojson.Basic.from_string(modelId);
+      let chunkId =
+        chunkId
+        |> Yojson.Basic.Util.to_list
+        |> List.hd
+        |> Yojson.Basic.Util.to_string;
```
davesnx, line 40: why is this necessary? modelId should be a string already

demo/server/server_actions/Route.re
```diff
@@ -0,0 +1,81 @@
...
+                String.length(form_prefix),
+                String.length(name) - String.length(form_prefix),
+              );
+            (key, value);
+          },
+          formData,
+        );
+      // Passing the formData as a hashtable to make it easier to handle as Dream don't
+      // have a better way to handle it
+      let formData = formData |> List.to_seq |> Hashtbl.of_seq;
```
davesnx, line 59: I see you treat FormData as a hashtable explicitly, do you think having a proper module called "FormData" could be helpful?

demo/server/server_actions/Route.re
```diff
@@ -0,0 +1,81 @@
...
+      let action = formDataManifest(actionId);
+      action(formData);
+    | _ =>
+      failwith(
+        "Missing $ACTION_ID, this formData was not created by server-reason-react",
+      )
+    }
+  | (Body(body), Some(actionId)) =>
+    let action = actionsManifest(actionId);
+    action(body);
```
davesnx, line 75: I'm not familiar with this branch, what's this case? Worth adding a comment?

packages/server-reason-react-ppx/cram/component-definition.t/input.re
```diff
@@ -68,16 +68,25 @@ module Upper_with_aria = {
   let make = (~children) => <div ariaHidden="true"> children </div>;
 };
 
-module Form_with_method = {
+// As the ppx does not runs as preprocessor, using method will break, like the error below:
+// +  File "output.ml", line 62, characters 33-41:
+// +  62 |     let make ~children  = ((form ~method:"GET" ~children:[children] ())
```
davesnx, line 73: `~method` is invalid in OCaml, since needs to be `~method_`. In Reason, this isn't in the best case where method is valid, when often OCaml keywords are invalid in Reason

packages/reactDom/test/test_RSC_model.ml
```diff
@@ -334,32 +334,6 @@ let client_with_element_props () =
...
-  in
-  let%lwt stream = ReactServerDOM.render_model (app ()) in
-  assert_stream stream
-    [
-      "2:I[\"./client-with-props.js\",[],\"ClientWithProps\"]\n";
-      "1:[[\"$\",\"div\",null,{\"children\":\"Server Content\"}],[\"$\",\"$2\",null,{\"promise\":\"$@3\"}]]\n";
-      "0:\"$1\"\n";
-      "3:\"||| Resolved |||\"\n";
-    ]
-
```
davesnx, line 362: Why delete this?

packages/reactDom/test/test_RSC_model.ml
```diff
@@ -434,6 +408,11 @@ let style_as_json () =
   assert_stream stream
     [ "0:[\"$\",\"div\",null,{\"style\":{\"zIndex\":\"34\",\"color\":\"red\",\"background\":\"blue\"}}]\n" ]
 
+let act_with_simple_response () =
+  let response = React.Json (`String "Server Content") in
+  let%lwt stream = ReactServerDOM.act response in
```
davesnx, line 413: Just curiosity, where does `act` comes from?

packages/react/src/React.ml
```diff
@@ -386,13 +390,14 @@ type element =
   | Consumer of element
   | Suspense of { key : string option; children : element; fallback : element }
 
-and client_props = (string * client_prop) list
+and client_props = (string * rsc_value) list
 
-and client_prop =
+and rsc_value =
```
davesnx, line 395: I would still call this client_prop, rsc_value doesn't add any meaning.

packages/reactDom/src/ReactDOM.ml
```diff
@@ -18,6 +18,11 @@ let attribute_to_html attr =
   (* true attributes render solely the attribute name *)
   | Bool (name, _, true) -> Html.present name
   | Style styles -> Html.attribute "style" (ReactDOMStyle.to_string styles)
+  (*
+     When we have a Action prop, means that we have a `server action` and not a `string action` in this prop, them we don't render it
+     because the server action will be provided as a model and react will handle the rest.
+  *)
+  | Action (_name, _, _) -> Html.omitted ()
```
davesnx, line 25: nit: can you make this comment like 1 line and 5 words?

```
server actions get ignored blabla
```

davesnx: also, I wonder if we need to raise when seeing one Action passed in ReactDOM, since actions in HTML do not make any sense

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -333,16 +362,17 @@ let rec to_html ~fiber (element : React.element) : (Html.element * json) Lwt.t =
...
                 let%lwt _html, model = to_html ~fiber element in
                 Lwt.return (name, model)
-            | Promise (promise, value_to_json) ->
+            | React.Promise (promise, value_to_json) ->
                 let context = Fiber.get_context fiber in
                 let _finished, parent_done = Lwt.wait () in
                 let index = Fiber.use_index fiber in
                 let sync = (name, `String (Model.promise_value index)) in
                 let async : Html.element Lwt.t =
+                  (* TODO: Add support for React.List, React.Assoc *)
```
davesnx, line 375: Remove this comment

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -404,14 +435,26 @@ let rec to_html ~fiber (element : React.element) : (Html.element * json) Lwt.t =
...
+    match chunk with
+    | Model.Chunk_value json ->
+        context.index <- id;
+        context.push (client_value_chunk_script id json)
+    | Model.Chunk_component_ref json ->
+        context.index <- id;
+        context.push (client_reference_chunk_script id json)
+  in
+  let context : Model.stream_context =
+    { push = push_chunk; close = context.close; chunk_id = context.index; pending = 0 }
```
davesnx, line 449: why creating a new context and not passing it via props?

davesnx (approved):
Last review and I think will be ready to merge, great work pedro!

packages/reactDom/test/test_RSC_model.ml
pedrobslisboa: It was deleted by mistake while cleaning the code. Thx

packages/react/src/React.ml
pedrobslisboa: But rsc_value is used for the server action return. `client_prop` is limited to a component. 
Could it be `client_value`?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -96,93 +102,95 @@ module Model = struct
...
+                let json = value_to_json value in
+                context.pending <- context.pending - 1;
+                context.push chunk_id (Chunk_value json);
+                if context.pending = 0 then context.close ();
+                Lwt.return ());
+            `String (promise_value chunk_id)
+        | Fail exn ->
+            (* TODO: Can we check if raise is good heres? *)
+            raise exn)
+    (* TODO: Implement Action for props ($T) *)
```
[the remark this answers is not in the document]
pedrobslisboa: Because we need to find a way to have func props on the Client component:

```
[@react.client.component]
let make =
    (
      ~noteId: option(int),
      ~children: React.element,
      ~onClick: option(_ => unit)=?,
    ) => {
```

I was thinking of creating it after this merge to not make this pull request to much complex

demo/server/server_actions/Route.re
pedrobslisboa: Options:

- Have a "collector" module to register and get the actions by ID. (this one looks more error-proof and easy to apply)
- Have a ppx that generates those manifest functions.

packages/reactDom/test/test_RSC_model.ml
pedrobslisboa: Todo: Change to act to the same name react uses on server to can action

packages/reactDom/src/ReactServerDOM.ml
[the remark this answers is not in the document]
pedrobslisboa: After talk: add this on the demo.

packages/react/src/React.ml
pedrobslisboa: Check for react name for this kind or use client_value

demo/universal/js/ReactServerDOMWebpack.re
```diff
@@ -0,0 +1,64 @@
...
+       let body = Fetch.BodyInit.make(body);
+       Fetch.fetchWithInit(
+         "/",
+         Fetch.RequestInit.make(~method_=Fetch.Post, ~headers, ~body, ()),
+       )
+       |> Js.Promise.then_(result => {
+            let body = Fetch.Response.body(result);
+            createFromReadableStream(body, ());
+          });
+     });
```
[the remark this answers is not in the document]
pedrobslisboa: TODO: Talk about the error handling

demo/universal/native/lib/Actions.re
pedrobslisboa: For this, we will need to check for the Error handling on react. We have a task for it on the board.

demo/server/server_actions/Route.re
pedrobslisboa: TODO: Create a FormData handler at server-reason-react.js

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: We discussed this on a call and agreed that we may need to improve the fiber and model contexts.
For now, this context will be as follows because we need to push to the chunk of the action.

demo/server/server_actions/Route.re
pedrobslisboa: React passes the model Id as a List: ["$K{ID}"].

To improve the read I'm changing to:
```reason
      let chunkId = Yojson.Basic.from_string(modelId);
      let chunkId = switch (chunkId) {
      | `List([`String(chunkId)]) => chunkId
      | _ => failwith("Invalid chunkId")
      };
```

demo/server/server_actions/Route.re
pedrobslisboa: https://github.com/ml-in-barcelona/server-reason-react/issues/228

demo/universal/native/lib/Actions.re
```diff
@@ -0,0 +1,408 @@
...
+    let log = (~request, message) => {
+      Dream.log("Logging: %s", message);
+      Lwt.return_none;
+    };
+  */
+  // It's going to be on top to this that we are going to generate the codes bellow
+
+  // As soon the user declare the action with a request, we pass it from the router to the handler,
+  [@platform native]
+  let logHandler = (~request, message) => {
```
pedrobslisboa, line 383: @davesnx I know that we talked about this, but I'm just adding a point here:

The user is going to declare:
```reason
  [@platform native]
  let logHandler = (~request, message) => {
    Dream.log("Logging: %s", message);
    Lwt.return_none;
  };
```

But it's going to be used like this:
```reason
[@react.client.component]
let make =
    (~log: [@react.server.action] (string => Js.Promise.t(unit))) => {
  <button
    className="font-mono border-2 py-1 px-2 rounded-lg bg-yellow-950 border-yellow-700 text-yellow-200 hover:bg-yellow-800"
    onClick=[%browser_only
      _ => {
        log("Hello server") |> ignore;
      }
    ]>
    {React.string("Click to log on server")}
  </button>;
};
```


As you can see here, the contract on action is `(~request, string) => Js.Promise.t(unit)` and on the usage is `(string) => Js.Promise.t(unit)`

Can it be confusing that the action contract we declare is not the same one we use?

I agree with the request as an arg. Just pointing it out.

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -765,6 +802,34 @@ let raise_usage_of_client_components module_expr =
         ]
   | _ -> module_expr
 
+(*
```
pedrobslisboa, line 805: @davesnx I created this transform to ensure props are declared with [react.server.action] that it will be `string` in the end as we pass IDs on Native instead of the function.
```
Native: React.Action(string)
Js: (...'a) => Js.Promise.t('b)
```

```reason
[@react.client.component]
let make =
    (~foo: [@react.server.action] (unit => Js.Promise.t(string))) => {

  //...
};

// Native
[@react.client.component]
let make =
    (~foo: [@react.server.action] string) => {

  //...
};

// Js
[@react.client.component]
let make =
    (~foo: [@react.server.action] (unit => Js.Promise.t(string))) => {

  //...
};
```
Do you have another idea for it?

demo/universal/native/lib/Actions.re
davesnx: I don't think we need the `~log: [@react.server.action]`

demo/server/actions/Notes.re
```diff
@@ -0,0 +1,64 @@
...
+
+  let title =
+    body |> Yojson.Basic.Util.member("title") |> Yojson.Basic.Util.to_string;
+  let content =
+    body |> Yojson.Basic.Util.member("content") |> Yojson.Basic.Util.to_string;
+  let note = DB.addNote(~title, ~content);
+  let%lwt response =
+    switch%lwt (note) {
+    | Ok(note) => Lwt.return(createNoteResponse(note))
+    | Error(e) => Lwt.return(React.RSC_value_Json(`String(e)))
```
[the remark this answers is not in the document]
pedrobslisboa: React has a different way of treating errors; they are handled as chunks. We should, in the future, have a React. Error

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -96,93 +102,95 @@ module Model = struct
...
-              | Fail exn ->
-                  (* TODO: Can we check if raise is good heres? *)
-                  raise exn))
-        props
-    in
-    let initial_chunk_id = get_chunk_id context in
-    context.push initial_chunk_id (Chunk_value (to_payload element));
-    if context.pending = 0 then context.close ()
+  let rec element_to_payload ~context element =
+    match (element : React.element) with
```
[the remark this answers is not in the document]
pedrobslisboa: @davesnx I think we cannot move the rec to an inner function because `client_values_to_json` is an `and` function that is used in `create_action_response`

pedrobslisboa:
As all the answers were asked, I took the action to merge it; any extra doubts must be solved into issues or the next PR.

PR #225, opened 2025-04-01
Feat/server action formdata

pedrobslisboa:
## Description

Introduces a new type of prop to handle action on Form as function


packages/reactDom/src/ReactDOM.ml
```diff
@@ -18,6 +18,8 @@ let attribute_to_html attr =
   (* true attributes render solely the attribute name *)
   | Bool (name, _, true) -> Html.present name
   | Style styles -> Html.attribute "style" (ReactDOMStyle.to_string styles)
+  (* ActionFunction should be always with empty value on the markup *)
```
davesnx, line 21: This comment doesn't add much, I would rather add comments that explain the why, not the how. Looking at L22 you can already guess that the output if empty.

packages/reactDom/src/ReactDOM.ml
```diff
@@ -18,6 +18,8 @@ let attribute_to_html attr =
...
+  | ActionFunction (name, _, _) -> Html.attribute name ""
```
davesnx, line 22: `Html.omitted ()` is the proper way to ignore attributes

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -35,19 +35,27 @@ module Model = struct
...
     match prop with
     (* We ignore the HTML name, and only use the JSX name *)
     | React.JSX.Bool (_, key, value) -> Some (key, `Bool value)
     (* We exclude 'key' from props, since it's outside of the props object *)
-    | React.JSX.String (_, key, _) when key = "key" -> None
+    | React.JSX.String (_, "key", _) -> None
+    (* We exclude 'method' from props, since on the client react is going to handle it when the form has actionFn *)
+    | React.JSX.String (_, "method", _) -> None
+    (* We exclude 'enctype' from props, since on the client react is going to handle it when the form has actionFn *)
+    | React.JSX.String (_, "enctype", _) -> None
```
davesnx, line 47: I don't understand those. Why do we remove them from the json payload?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -102,8 +110,14 @@ module Model = struct
...
-        let props = props_to_json attributes in
-        node ~key ~tag ~props (List.map (element_to_payload ~context) children)
+        let is_input_action_input =
+          tag = "input"
+          && List.exists
+               (fun attr -> match attr with React.JSX.String (_, "name", name) -> name = "$ACTION_ID" | _ -> false)
+               attributes
+        in
+        let props = props_to_json ~context attributes in
+        if is_input_action_input then `Null else node ~key ~tag ~props (List.map (element_to_payload ~context) children)
```
davesnx, line 120: Why?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -35,19 +35,27 @@ module Model = struct
...
-  let prop_to_json (prop : React.JSX.prop) =
+  let prop_to_json ~context (prop : React.JSX.prop) =
     match prop with
     (* We ignore the HTML name, and only use the JSX name *)
     | React.JSX.Bool (_, key, value) -> Some (key, `Bool value)
     (* We exclude 'key' from props, since it's outside of the props object *)
-    | React.JSX.String (_, key, _) when key = "key" -> None
+    | React.JSX.String (_, "key", _) -> None
+    (* We exclude 'method' from props, since on the client react is going to handle it when the form has actionFn *)
+    | React.JSX.String (_, "method", _) -> None
```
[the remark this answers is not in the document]
pedrobslisboa: Yep, I forgot to treat it

pedrobslisboa:
ARCHIVE

PR #229, opened 2025-04-10
Minimum for server actions demo

pedrobslisboa:
## Description

Create the minimum for this demo without new ppxs and minimum magic work. 
This PR will count with the demo related to:

- Declare actions
    - For the server: router and action_id
    - For the client: request function
- Execute those actions on the client with react tools

Todos

- [x]  Handle actions that are imported on client components
- [x]  Adapt ReactServerDOM to be able to handle functions.
- [x]  Create action samples showing it decoding the args


packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -333,16 +360,17 @@ let rec to_html ~fiber (element : React.element) : (Html.element * json) Lwt.t =
...
                 let%lwt _html, model = to_html ~fiber element in
                 Lwt.return (name, model)
-            | Promise (promise, value_to_json) ->
+            | React.Promise (promise, value_to_json) ->
                 let context = Fiber.get_context fiber in
                 let _finished, parent_done = Lwt.wait () in
                 let index = Fiber.use_index fiber in
                 let sync = (name, `String (Model.promise_value index)) in
                 let async : Html.element Lwt.t =
+                  (* TODO: Add support for React.List, React.Assoc *)
```
pedrobslisboa, line 373: Delete this

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -484,6 +484,10 @@ let get_labelled_arguments pvb_expr =
 
 let make_of_json ~loc (core_type : core_type) prop =
   match core_type with
+  (* QUESTION: How do we handle especial types on props, 
+     like `("someProp"), `List([React.element, string]). 
+     We already support it, but not with the ppx. 
+     Checkout the test_RSC_model.ml for more details. packages/reactDom/test/test_RSC_html.ml *)
```
pedrobslisboa, line 490: Delete this

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -96,93 +96,98 @@ module Model = struct
...
-                  (name, `String (promise_value chunk_id))
-              | Fail exn ->
-                  (* TODO: Can we check if raise is good heres? *)
-                  raise exn))
-        props
-    in
-    let initial_chunk_id = get_chunk_id context in
-    context.push initial_chunk_id (Chunk_value (to_payload element));
-    if context.pending = 0 then context.close ()
+  let rec element_to_payload ~context element =
```
pedrobslisboa, line 99: @davesnx I think we cannot move the rec to an inner function because `client_values_to_json` is an `and` function that is used in `createActionResponse`

PR #232, opened 2025-04-15
Change actions to uncurry

pedrobslisboa:
## Description

This improves the code by leaving the args uncurry, avoiding having the args as a single value in the action function.


demo/server/server_actions/Route.re
```diff
@@ -24,9 +24,10 @@ let formDataManifest = (id: string) => {
 [@platform native]
 let getArgs = body => {
   switch (Yojson.Basic.from_string(body)) {
-  | `List([`List(args)]) => args
   // When there is no args, the react will send a list with a single string "$undefined"
+  // TODO: Create a Helper to "parse" special values like "$undefined"
```
davesnx, line 28: Do you think we need the opposite as the render_model, like parse_model?

davesnx (approved):

PR #234, opened 2025-04-25
Use tailwind from the CLI

davesnx:


packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -570,6 +570,15 @@ let render_html ?(debug = false) ?(bootstrapScriptContent = "") ?(bootstrapScrip
...
+  let html_bootstrap_stylesheets =
+    match bootstrapStylesheets with
+    | [] -> Html.null
+    | stylesheets ->
+        stylesheets
+        |> List.map (fun stylesheet ->
+               Html.node "link" [ Html.attribute "href" stylesheet; Html.attribute "rel" "stylesheet" ] [])
+        |> Html.list
+  in
   let head =
```
davesnx, line 582: while looking into this, I found out that we are rendering `<body />` mega wrong. cc @pedrobslisboa 

will look into this soon

PR #241, opened 2025-05-05
Extract client modules

pedrobslisboa:
## Description

Add the support to Server Functions on `extract-client-modules` (former `extract-client-components`)

The idea is to change the extractor and make it possible to extract server functions, similar to what we do with the client components. That's why I change the tool name to extract-client-modules (@davesnx don't know if it's a good name)

You can check the behavior by seeing the test or checking the `window.__server_functions_manifest_map` on demo:
Captura de Tela 2025-05-05 às 19 12 07

### DIFF
--- a/demo/universal/js/ReactServerDOMEsbuild.js
+++ b/demo/universal/js/ReactServerDOMEsbuild.js
@@ -113,6 +113,7 @@ const ReactFlightClientConfigBundlerEsbuild = {
     debug("resolveClientReference", bundlerConfig, metadata);
     // Reference is already resolved during the build
     return {
+      type: "ClientComponent",
       id: metadata[ID],
       name: metadata[NAME],
       bundles: metadata[BUNDLES],
@@ -120,18 +121,11 @@ const ReactFlightClientConfigBundlerEsbuild = {
   },
 
   resolveServerReference(bundlerConfig, ref) {
-    debug("resolveServerReference", ref);
-    const idx = ref.lastIndexOf("#");
-    const id = ref.slice(0, idx);
-    const name = ref.slice(idx + 1);
-    const bundles = bundlerConfig[id];
-    if (!bundles) {
-      throw new Error("Invalid server action: " + ref);
-    }
+    debug("resolveServerReference", bundlerConfig, ref);
+
     return {
-      id,
-      name,
-      bundles,
+      type: "ServerFunction",
+      id: ref,
     };
   },
 
@@ -142,14 +136,25 @@ const ReactFlightClientConfigBundlerEsbuild = {
   },
 
   requireModule(metadata) {
-    const component = window.__client_manifest_map[metadata.id];
-    if (!component) {
-      /* TODO: use reportGlobalError? */
-      throw new Error(
-        `Could not find client component with id: ${metadata.id}`
-      );
+    const getModule = (type, id) => {
+      switch (type) {
+        case "ServerFunction":
+          const fn = window.__server_functions_manifest_map[id];
+
+          return fn;
+        case "ClientComponent":
+          const component = window.__client_manifest_map[id];
+
+          return component
+      }
     }
-    return component;
+
+    const module = getModule(metadata.type, metadata.id);
+    if (!module) {
+      throw new Error(`Could not find module of type ${metadata.type} with id: ${metadata.id}`);
+    }
+
+    return module
   },
 };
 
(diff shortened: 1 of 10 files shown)


packages/server-reason-react-ppx/cram/client-component-e2e.t/run.t
```diff
@@ -91,7 +91,7 @@
                         module J = {
                           [@ocaml.warning "-unboxable-type-in-prim-decl"]
                           [@mel.internal.ffi
-                            "����\000\000\000\011\000\000\000\005\000\000\000\r\000\000\000\012���A�$name@"
+                            "����\000\000\000\011\000\000\000\005\000\000\000\r\000\000\000\012���A�$name@"
```
pedrobslisboa, line 94: @davesnx There is something strange here. I made no changes to this file besides the one at line 20, as you can see on the diff below.

Captura de Tela 2025-05-05 às 19 42 57

Do you know what is happening?
I tested and saw that when I changed the name of the tool from extract-client-components to `extract-client-modules`, it started to happen.

demo/universal/native/lib/ServerFunctions.re
```diff
@@ -248,6 +281,20 @@ module Samples = {
...
+  | Server => ()
+  | Client =>
+    // Comment to esbuild create the manifest
+    [%mel.raw
+     {|
+    // extract-server-function id/samples/simpleResponse simpleResponse Samples
+    ''
+    |}
+    ]
+  };
```
pedrobslisboa, line 297: While we don't have the ppx, I have to add it with the switch platform, with a little hack adding an empty string cause only `%mel.raw` was working, not the `%%mel.raw`

demo/universal/native/lib/ServerFunctions.re
```diff
@@ -66,6 +66,17 @@ module Notes = {
...
+  // This is a little hack while we don't have the ppx to generate the comment
+  switch%platform () {
+  | Server => ()
+  | Client => [%mel.raw
+     {|
+    // extract-server-function id/notes/create create Notes
+    ''
+    |}
+    ]
+  };
```
pedrobslisboa, line 79: While we don't have the ppx, I have to add it with the switch platform, with a little hack adding an empty string cause only `%mel.raw` was working, not the `%%mel.raw`.

In the end, melange will not add the empty string

packages/server-reason-react-ppx/cram/client-component-e2e.t/run.t
pedrobslisboa: You can see the error here as well: https://github.com/ml-in-barcelona/server-reason-react/actions/runs/14847718294/job/41685229888

As you can see in the image, the test has a strange fail:
Captura de Tela 2025-05-05 às 23 28 56

davesnx (approved):
Looks correct, I would investigate the need for window.xxx since It might not be needed.

I would say renaming a module and adding functionality in the same PR is a bad idea, since it's not possible to review the diff. Also, if we don't have a clear name in mind, I wouldn't bother a rename like this

demo/universal/js/ReactServerDOMEsbuild.js
```diff
@@ -142,14 +138,25 @@ const ReactFlightClientConfigBundlerEsbuild = {
   },
 
   requireModule(metadata) {
-    const component = window.__client_manifest_map[metadata.id];
-    if (!component) {
-      /* TODO: use reportGlobalError? */
```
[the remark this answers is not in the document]
pedrobslisboa: Ops, sorry, miss-deleted

pedrobslisboa:
> I would say renaming a module and adding functionality in the same PR is a bad idea, since it's not possible to review the diff. Also, if we don't have a clear name in mind, I wouldn't bother a rename like this

I also revert the renaming.
In the future we can think more about it.

demo/universal/js/ReactServerDOMEsbuild.js
```diff
@@ -113,25 +113,21 @@ const ReactFlightClientConfigBundlerEsbuild = {
     debug("resolveClientReference", bundlerConfig, metadata);
     // Reference is already resolved during the build
     return {
+      type: "ClientComponent",
```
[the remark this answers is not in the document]
pedrobslisboa: Nope, on other plugins, they just treat all modules with the same contract: `[id, name,chunks]`

As we need to differentiate the client components from the server functions, I added it.

PR #258, opened 2025-05-27
Initial work for server function ppx

pedrobslisboa:
## Description

This is the initial work for the server function ppx, it does not cover yet FormData, only working with Body content.

## Details

To declare a server function you must use the `@react.server.function` attribute:
```reason
[@react.server.function]
let simpleResponse = (~name: string, age: int): Js.Promise.t(string) => {
  Lwt.return(Printf.sprintf("Hello %s, you are %d years old", name, age));
};
```

### Server

The result of this code on server will be:
```reason
[@react.server.function]
let simpleResponse = {
  Runtime.id: "111391064",
  call: (~name: string, ~age: int) => (
    Lwt.return(
      Printf.sprintf("Hello %s, you are %d years old", name, age),
    ):
      Js.Promise.t(string)
  ),
};
let simpleResponseRouteHandler = args => {
  let (name, age) =
    switch (args) {
    | [name, age] =>
      let name = string_of_json(name);
      let age = int_of_json(age);
      (name, age);
    | _ =>
      failwith(
        "server-reason-react: invalid arguments",
      )
    };
  try(
    simpleResponse.call(~name, ~age)
    |> Lwt.map(response => React.Json(string_to_json(response)))
  ) {
  | e => Lwt.fail(e)
  };
};
```

- `simpleResponse` turns into the server_function contract: `{id;call}`
- `simpleResponseRouteHandler` is the function the dev is going to use when the action router is called
- As you can see, we does the decode and encode of the args and response (For now, server functions only allows to response JSON and the args are limited to json-like values.)

### Client

The result of the code on client will be:

```reason
module J = {
  [@ocaml.warning "-unboxable-type-in-prim-decl"]
  external unsafe_expr: _ => _ = "#raw_stmt";
};
J.unsafe_expr(
  "// extract-server-function 500511824 simpleResponse ",
);
};
[@react.server.function]
let simpleResponse = {
  Runtime.id: "500511824",
  call: (~name: string, age: int) => {
    let action =
      ReactServerDOMEsbuild.createServerReference("500511824");
    (
      [@ocaml.warning "-ignored-extra-argument"]
      Js.Internal.opaqueFullApply(
        (Js.Internal.opaque((action: Js.Fn.arity2(_)).I2))(
          ~name,
          age,
        ),
      ): _
    );
  },
};
```
- `simpleResponse` turns into the server_function contract: `{id;call}`
- We create a comment to identify the server function on the esbuild: `// extract-server-function 500511824 simpleResponse`

## Question

@davesnx I didn't know at the time if the server-reason-react.ppx was the right place, but following the idea of "make it first and improve later", I put all the code well splitted to be easy to unplug from there if necessary.

## After the ppx, ServerFunctions.re dropped from 458 lines to 112


packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -847,7 +849,224 @@ let rewrite_jsx =
...
+              [%expr React.Json [%e json]]
+          | Some _ -> [%expr [%ocaml.error "server-reason-react: server functions must return a promise"]]
+          | _ ->
+              [%expr
+                [%ocaml.error "server-reason-react: server functions must have a return type annotation (Js.Promise.t)"]]
+          )
+    in
+    aux expr None
+
+  let map_arguments_to_expressions ~loc _name args =
```
pedrobslisboa, line 897: Why _name is ignored here?
```suggestion
  let map_arguments_to_expressions ~loc args =
```

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -847,7 +849,224 @@ let rewrite_jsx =
...
+    | _ ->
+        Some
+          (value_binding ~loc
+             ~pat:[%pat? [%p ppat_tuple ~loc (List.map ~f:(fun (_, label, _) -> pvar ~loc label) args_to_decode)]]
+             ~expr:
+               [%expr
+                 match args with
+                 | [%p plist ~loc (List.map ~f:(fun (_, label, _) -> pvar ~loc label) args_to_decode)] ->
+                     [%e decoded_args args]
+                 | _ -> failwith "server-reason-react: client components need type annotations"])
```
pedrobslisboa, line 940: We could always return the value_binding and make the validation of empty list on the usage of it:
```suggestion
        Some
          (value_binding ~loc
             ~pat:[%pat? [%p ppat_tuple ~loc (List.map ~f:(fun (_, label, _) -> pvar ~loc label) args_to_decode)]]
             ~expr:
               [%expr
                 match args with
                 | [%p plist ~loc (List.map ~f:(fun (_, label, _) -> pvar ~loc label) args_to_decode)] ->
                     [%e decoded_args args]
                 | _ -> failwith "server-reason-react: client components need type annotations"])
```

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -847,7 +849,224 @@ let rewrite_jsx =
...
+
+  let create_router_handler ~loc name args make_response =
```
pedrobslisboa, line 942: This make_response is complex and does not mean too much. Please change it to response_core_type and handle it internally

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -847,7 +849,212 @@ let rewrite_jsx =
...
+            [%e decoded_args args]]
+    in
+    value_binding ~loc
+      ~pat:[%pat? [%p ppat_tuple ~loc (List.map ~f:(fun (_, label, _) -> pvar ~loc label) args_to_decode)]]
+      ~expr:
+        [%expr
+          match args with
+          | [%p plist ~loc (List.map ~f:(fun (_, label, _) -> pvar ~loc label) args_to_decode)] ->
+              [%e decoded_args args_to_decode]
+          | _ -> failwith "server-reason-react: invalid arguments"]
```
pedrobslisboa, line 936: Can we do it individually instead of all together on a single switch?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -854,7 +856,213 @@ let rewrite_jsx =
...
+          include struct
+            [%%i extract_client_raw]
+            [%%i pstr_value ~loc:structure_item.pstr_loc rec_flag [ value_binding ]]
+          end]
+    | _ -> structure_item
+
+  let traverse =
+    object (_)
+      inherit [Expansion_context.Base.t] Ppxlib.Ast_traverse.map_with_context as super
+      val mutable nested_modules = []
```
pedrobslisboa: @davesnx I saw how you did for sub-modules, but do you think we should have it globally? When I create the server functions ppx, I set it as a mutable value inside the traverse object.

Having it globally looks strange to me and makes it self-contained to the object, and its context looks clearer.

davesnx: it doesn't really matter I would say, part of the traverse make sense as well

pedrobslisboa: Yeah, in the end, it indeed does not matter. My question is more related to the following:
If the value is "global," but the method to assign sub_modules is inside `rewrite_jsx,` it looks strange to me that a `sub_module` is global and dependent on a specific traverse object.

packages/server-reason-react-ppx/cram/server-function-on-server.t/input.re
```diff
@@ -0,0 +1,30 @@
+include Melange_json.Primitives;
```
davesnx: open right?

packages/server-reason-react-ppx/cram/server-function-on-server.t/input.re
```diff
@@ -0,0 +1,30 @@
...
+
+[@react.server.function]
+let withOptionalArg = (~name: string="Lola", ()): Js.Promise.t(string) => {
+  Lwt.return(Printf.sprintf("Hello, %s", name));
+};
+
+[@react.server.function]
+let withNoArgs = (): Js.Promise.t(string) => {
+  Lwt.return("Hello, world!");
+};
```
davesnx, line 45: I feel there's a few cases uncovered:

- a unlabelled arguments only?
- optional argument?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -854,7 +856,213 @@ let rewrite_jsx =
...
+        in
+        let core_type_string = string_of_core_type core_type in
+        let of_json = make_of_json ~loc core_type [%expr args.([%e eint ~loc i])] in
+        value_binding ~loc
+          ~pat:[%pat? [%p ppat_var ~loc { txt = label; loc }]]
+          ~expr:
+            [%expr
+              try [%e of_json]
+              with _ ->
+                failwith
```
davesnx: Why would you failwith, and not raise an Invalid_argument? 

Also, where the try/catch above to capture this?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -854,7 +856,213 @@ let rewrite_jsx =
...
+                  (Printf.sprintf "server-reason-react: error on decoding argument '%s'. EXPECTED: %s, RECEIVED: %s"
```
davesnx: This error is rare, I would consider have completly different errors in development and production.

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -854,7 +856,213 @@ let rewrite_jsx =
...
+    let body_expr =
+      match args_to_decode with
+      | [] -> encoded_response_expr
+      | args_to_decode ->
+          let decoded_expr = decode_arguments_vb ~loc args_to_decode in
+          pexp_let ~loc Nonrecursive decoded_expr encoded_response_expr
+    in
+    [%stri FunctionReferences.register [%e estring ~loc id] (Body (fun args -> [%e body_expr]))]
+
+  let create_server_function_contract ~loc id expression =
```
davesnx: nit: `contract` is a weird name for OCaml, since there's no interfaces, neither traits.

I know what you mean, but would suggest create_server_function_record or type

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -854,7 +856,213 @@ let rewrite_jsx =
...
+        let value_binding = { vb with pvb_expr = create_server_function_contract ~loc:vb.pvb_loc id base_fn } in
+        let stri =
+          [%stri
+            include struct
+              [%%i pstr_value ~loc rec_flag [ value_binding ]]
+              [%%i create_router_handler ~loc ~id ~function_name ~args ~core_type:return_core_type]
+            end]
+        in
+        stri
+    | _ -> structure_item
```
davesnx: Does multiple `let xxx and` work? Can we raise if it doesn't? or support if it's trivial?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -854,7 +856,213 @@ let rewrite_jsx =
...
+
+  let create_client_function ~loc id args =
+    let apply_args = map_arguments_to_expressions ~loc args |> List.map ~f:(fun (_, expr) -> (Nolabel, expr)) in
+    let fn =
+      [%expr
+        (* TODO: support others bundlers? *)
```
davesnx: Currently we only support our esbuild, and there's no plans on supporting Webpack at the moment (or any other really)

packages/server-reason-react-ppx/server_reason_react_ppx.ml
davesnx: Yeah, this change make sense

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -854,7 +856,213 @@ let rewrite_jsx =
...
+    end
+end
+
 let () =
   Driver.add_arg "-melange" (Unit (fun () -> mode := Js)) ~doc:"preprocess for js build";
-  Ppxlib.Driver.V2.register_transformation "server-reason-react.ppx" ~preprocess_impl:rewrite_jsx#structure
+  Ppxlib.Driver.V2.register_transformation "server-reason-react.ppx"
+    ~preprocess_impl:(fun ctx str ->
+      let str = rewrite_jsx#structure ctx str in
+      let str = ServerFunction.traverse#structure ctx str in
```
davesnx: why did you choose to create a new traverse, instead of adding everything in rewrite_jsx? (and rename it?)

I remember you mention it in a meeting, but now we are doing the double of iterations over an AST.

packages/server-reason-react-ppx/test/test.re
```diff
@@ -428,9 +464,85 @@ let multiple_contexts = () => {
...
+Alcotest_lwt.run(
   "server-reason-react.ppx",
   [
+    test_lwt("server_function_reference", server_function_reference),
+    test_lwt(
+      "server_function_reference_args_error",
+      server_function_reference_args_error,
+    ),
+    test_lwt("server_function", server_function),
+    test_lwt("server_function_curried", server_function_curried),
```
davesnx: What are we really testing here?

packages/server-reason-react-ppx/cram/server-function-on-server.t/input.re
pedrobslisboa: Yep, I added the optional arg here: https://github.com/ml-in-barcelona/server-reason-react/pull/258/files/caddd11e6b429585fb47211fae7f780754662dc7#diff-b93aac6ec9f405bec31d4f648cf16b4c7f8e61b4283b6da3b4d72189c4042f1aR22-R25

But It's with default value, I'm going to add one without the default value

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: I just left it as a reminder. Should I delete it?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: I split it until we were sure this was the place of the server function ppx. If it's good here, I will join it to the `rewrite_jsx` (renaming it).

packages/server-reason-react-ppx/test/test.re
pedrobslisboa: Ops, this was one test that I added when we were talking about the formData and bind, and I forgot to clean it.

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: > Why would you failwith, and not raise an Invalid_argument?

True, I'm going to update it. ~Probably it should also fail with `Lwt.fail`, right?~


> Also, where the try/catch above to capture this?

What do you mean by that? In our case, ReactServerDOM will propagate and catch the error.

packages/server-reason-react-ppx/server_reason_react_ppx.ml
davesnx: But we are sure, no?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
davesnx: I think so, yeah

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: I was sure, but we had a meeting a long time ago about "what belongs to where", and as I didn't remember if server function ppx was related to this talk I decided to push it and modify the review.

I will refactor it to fit well now.

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: TBH, I didn't try it. I'll check if the support is trivial. If not, I will raise it.

packages/server-reason-react-ppx/cram/client-component-on-the-client-nested.t/run.t
```diff
@@ -76,7 +76,7 @@
       type lola = {name: string};
   
       include {
-                [%%raw "// extract-client input.re Very_nested.Inner"];
```
pedrobslisboa, line 79: @davesnx Was it right? With the `Very_nested` at the beginning?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -31,6 +30,7 @@ let make_string ~loc str = Ast_helper.Exp.constant ~loc (Ast_helper.Const.string
 let react_dot_component = "react.component"
 let react_dot_async_dot_component = "react.async.component"
 let react_dot_client_dot_component = "react.client.component"
+let react_dot_server_function = "react.server.function"
```
davesnx: react_dot_server_dot_function

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -696,7 +709,183 @@ let props_to_model ~loc (props : (arg_label * expression option * pattern) list)
...
+           let core_type_string = string_of_core_type core_type in
+           let of_json = make_of_json ~loc core_type [%expr args.([%e eint ~loc i])] in
+           value_binding ~loc
+             ~pat:[%pat? [%p ppat_var ~loc { txt = label; loc }]]
+             ~expr:
+               [%expr
+                 try [%e of_json]
+                 with _ -> (
+                   match Sys.getenv_opt "ENV" with
+                   | Some "development" ->
```
davesnx: Wait, this is not correct. The env matters at runtime, not at build time. We pass `~env` on ReactServerDOM, we should use that.

Don't stress on this, keep the error on development for now, and we will revisit this in the future, add a todo please.

packages/server-reason-react-ppx/cram/client-component-on-the-client-nested.t/run.t
davesnx: 😭 good catch

demo/universal/native/shared/ServerFunctions.re
```diff
@@ -1,34 +1,8 @@
...
-    Note.(
-      `Assoc([
-        ("id", `Int(note.id)),
-        ("title", `String(note.title)),
-        ("content", `String(note.content)),
-        ("updated_at", `Float(note.updated_at)),
-      ])
-    );
-  };
+include Melange_json.Primitives;
```
davesnx: It's an open, right?

demo/universal/native/shared/ServerFunctions.re
```diff
@@ -128,282 +25,28 @@ module Notes = {
...
-      Runtime.id: errorId,
-      call: _ => {
-        let action = ReactServerDOMEsbuild.createServerReference(errorId);
-        action(.);
-      },
-    }
-  };
-
-let formDataId = "id/formData";
+let formDataId = "id/samples/formData";
```
davesnx, line 51: is this Id needed?

demo/universal/native/shared/ServerFunctions.re
```diff
@@ -128,282 +25,28 @@ module Notes = {
...
-
-[@platform native]
-let errorHandler = () => {
+[@react.server.function]
+let error = (): Js.Promise.t(string) => {
   // Uncomment to see that it also works with Lwt.fail
   // Lwt.fail(failwith("Error from server"));
-  Lwt.fail(
-    failwith("Error from server"),
+  failwith(
```
davesnx: Why failwith here, when before was a Lwt.fail? (also, failwith is banned here)

davesnx: failwith lies to the type-checker, the function has `Js.Promise.t(string)` while in reality is crashing hard

demo/universal/native/shared/ServerFunctions.re
```diff
@@ -447,3 +84,6 @@ let formData =
       },
     }
   };
+
+[@platform native]
+FunctionReferences.register(formData.id, FormData(formDataRouteHandler));
```
davesnx, line 91: This should not be here?

packages/server-reason-react-ppx/cram/server-function.t/run.t
```diff
@@ -0,0 +1,182 @@
...
+            {
+              module J = {
+                [@ocaml.warning "-unboxable-type-in-prim-decl"]
+                external unsafe_expr: _ => _ = "#raw_stmt";
+              };
+              J.unsafe_expr(
+                "// extract-server-function 1073617701 withLabelledArg ",
+              );
+            };
+            [@react.server.function]
```
davesnx: The attribute should not be present after the ppx

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -696,7 +709,183 @@ let props_to_model ~loc (props : (arg_label * expression option * pattern) list)
...
+        pexp_fun ~loc arg_label arg_expression fun_pattern (last_expr_to_fn ~loc expression fn)
+    | _ -> fn
+
+  let generate_id ~loc name =
+    (* We need to add a nasty hack here, since have different files for native and melange.Assume that the file structure is native/lib and js, and replace the name directly. This is supposed to be temporal, until dune implements https://github.com/ocaml/dune/issues/10630 *)
+    let file_path = loc.loc_start.pos_fname |> Str.replace_first (Str.regexp {|/js/|}) "/native/shared/" in
+    let hash = Printf.sprintf "%s_%s_%d" name file_path loc.loc_start.pos_lnum |> Hashtbl.hash |> string_of_int in
+    hash
+
+  let get_arg_details (arg : arg_label * expression option * pattern) =
```
davesnx, line 726: what does this function do? 

Resolve patterns into the core_type, but also validates stuff? I'm confused by this abstraction

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -714,22 +903,25 @@ let rewrite_structure_item structure_item =
              in JavaScript, in the server, that doesn't make sense. If you need to render this on the server, \
              implement a placeholder or an empty element"]])
   (* let make = ... *)
+  | Pstr_value (rec_flag, value_bindings) when isReactServerFunctionBinding (List.hd value_bindings) ->
+      let vb = List.hd value_bindings in
+      ServerFunction.rewrite_native_function ~vb ~rec_flag structure_item
```
davesnx: if you move this function below, you might get `and` for free

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -786,28 +977,21 @@ let rewrite_structure_item_for_js ctx structure_item =
...
-      match structure_item.pstr_desc with
-      | Pstr_value (_, value_bindings) -> List.exists ~f:isReactClientComponentBinding value_bindings
-      | _ -> false)
-    structure
-
-let rewrite_jsx =
+let traverse =
   object (_)
     inherit [Expansion_context.Base.t] Ppxlib.Ast_traverse.map_with_context as super
+    val mutable sub_modules = []
```
davesnx: can you name it `nested_module_names`?

demo/universal/native/shared/ServerFunctions.re
pedrobslisboa: I changed it so people can see we can fail with `raise` and `Lwt.fail`; both work. As you can see in the comment:
```reason
  // Uncomment to see that it also works with Lwt.fail
  // Lwt.fail(failwith("Error from server"));
  failwith(
    "Error from server",
  );
```

I've made the error handler work with both `raise` and `Lwt.fail`, so for us, it does not matter how the user raises it.
I also allowed this, due to this comment on Lwt.fail:
> Lwt.fail exn is like Lwt.return, except the new promise that is already rejected with exn.
Whenever possible, it is recommended to use raise exn instead, as raise captures a backtrace, while Lwt.fail does not. If you call raise exn in a callback that is expected by Lwt to return a promise, Lwt will automatically wrap exn in a rejected promise, but the backtrace will have been recorded by the OCaml runtime.
-----
> also, failwith is banned here

That is why I think `failwith` or `raise` should not be banned from being used on an Lwt function.

demo/universal/native/shared/ServerFunctions.re
pedrobslisboa: This PR is Body only:
> This is the initial work for the server function ppx, it does not cover yet FormData, only working with Body content.

I'm making it with small PR and context, to be easier to review.

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: It gets the args details: the arg label, name (from Labelled and Unlabel), and core_type.
If the arg doesn't fit on what expect it throws. It is a simpler version of [props_of_model](https://github.com/ml-in-barcelona/server-reason-react/blob/b3bb47be4ce56bcc2269d81d3d9b74bf50f07243/packages/server-reason-react-ppx/server_reason_react_ppx.ml#L553-L583)

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: By the way, we also treat the environment and error inside the ReactServerDOM. We don't need to think this is an dev error here on the ppx:
https://github.com/ml-in-barcelona/server-reason-react/blob/3241bb3343c2ec910b522129414ffa383f69169e/packages/reactDom/src/ReactServerDOM.ml#L74-L83

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: `ServerFunction.rewrite_native_function` returns a `stri`, not a `value_binding`.

I can change it to support the `and`, or we can let it for a future polish.

PR #259, opened 2025-05-30
feat: create function references at ReactServerDOM

pedrobslisboa:
## Description

This PR takes advantage of the fact that the "FunctionsRegistry" doesn't need to be included anywhere to be included as part of the ReactServerDOM.

It also incorporate the `decodeFormDataReply` to `ReactServerDOM`


demo/server/DreamRSC.re
```diff
@@ -55,34 +55,95 @@ let stream_html =
   );
 };
 
-let stream_server_action = fn => {
+let handleFormRequest = (actionId, formData) => {
+  // For now we're using hashtbl for FormData as we still cannot support the Js.FormData.t.
+  let formData =
+    formData
+    |> List.fold_left(
```
davesnx: nit: List.iter might be just more correct

packages/reactDom/src/ReactServerDOM.mli
```diff
@@ -13,8 +13,16 @@ val render_model :
 val create_action_response :
   ?env:[ `Dev | `Prod ] -> ?debug:bool -> ?subscribe:(string -> unit Lwt.t) -> React.client_value Lwt.t -> unit Lwt.t
 
-type server_function =
-  | FormData of (Js.FormData.t -> React.client_value Lwt.t)
-  | Body of (Yojson.Basic.t list -> React.client_value Lwt.t)
+module FunctionReferences : sig
+  type server_function =
+    | FormData of (Js.FormData.t -> React.client_value Lwt.t)
+    | Body of (Yojson.Basic.t array -> React.client_value Lwt.t)
```
davesnx: I would keep the type and decodeReply outside of "FunctionReferences"

Also, the registry is part of DreamRSC in my mental model imho

demo/server/DreamRSC.re
pedrobslisboa: To me, it is just a different approach. Is there a reason `List.iter` is better than `List.fold`?

```reason
  let formDataJs = Js.FormData.make();
  formData
    |> List.iter(
         ((name, value)) => {
           // For now we're only supporting strings.
           let (_filename, value) = value |> List.hd;
           Js.FormData.append(formDataJs, name, `String(value));
         },
       );
```

```reason
  let formData =
    formData
    |> List.fold_left(
         (acc, (name, value)) => {
           // For now we're only supporting strings.
           let (_filename, value) = value |> List.hd;
           Js.FormData.append(acc, name, `String(value));
           acc;
         },
         Js.FormData.make(),
       );
```

pedrobslisboa: As we talked on a private channel. The approach of having `List.iter` would be more performatic.

packages/reactDom/src/ReactServerDOM.mli
```diff
@@ -15,6 +15,23 @@ val create_action_response :
 
 type server_function =
   | FormData of (Js.FormData.t -> React.client_value Lwt.t)
-  | Body of (Yojson.Basic.t list -> React.client_value Lwt.t)
+  | Body of (Yojson.Basic.t array -> React.client_value Lwt.t)
 
-val decodeReply : string -> Yojson.Basic.t list
+val decodeReply : string -> Yojson.Basic.t array
```
davesnx, line 20: why move from array to list?

demo/universal/native/FunctionReferences.re
```diff
@@ -0,0 +1,72 @@
+include ReactServerDOM.FunctionReferencesMake({
```
davesnx: I think functionReferences could not be a functor, it doesn't add anything.

pedrobslisboa: How do we enforce the contract for `FunctionReferences`?

packages/reactDom/src/ReactServerDOM.mli
pedrobslisboa: Sorry, I should comment about it. It's more relate to a easy way to generate the decode of it:
https://github.com/ml-in-barcelona/server-reason-react/pull/258/files#diff-1c07547590d33ece4b45936aac75b29ecb883f7d039aaa169e949971ae656ac8R58-R82
```reason
                  let name =
                    try(string_of_json(args[0])) {
                    | _ =>
                      failwith(
                        Printf.sprintf(
                          "server-reason-react: error on decoding argument '%s'. EXPECTED: %s, RECEIVED: %s",
                          "name",
                          "string",
                          args[0] |> Yojson.Basic.to_string,
                        ),
                      )
                    }
                  and age =
                    try(int_of_json(args[1])) {
                    | _ =>
                      failwith(
                        Printf.sprintf(
                          "server-reason-react: error on decoding argument '%s'. EXPECTED: %s, RECEIVED: %s",
                          "age",
                          "int",
                          args[1] |> Yojson.Basic.to_string,
                        ),
                      )
                    };
```

PR #263, opened 2025-06-04
Add server function to action

pedrobslisboa:
## Description

 This PR makes the `action` attribute work with `string` and `function`.

## Details

The action attribute now works with`` `String of string | `Function of 'a Runtime.server_function ``.

> [!IMPORTANT]  
> This makes action incompatible with reason-react. It requires the usage of switch%platform always.

### DIFF
--- a/demo/universal/native/shared/ServerActionWithFormDataServer.re
+++ b/demo/universal/native/shared/ServerActionWithFormDataServer.re
@@ -1,10 +1,11 @@
-[@react.client.component]
+[@react.component]
 let make = () => {
   <form
     action={
       switch%platform () {
-      | Server => `String("")
-      | Client => Obj.magic(ServerFunctions.formData.call)
+      | Server => `Function(ServerFunctions.formData)
+      // doesn't matter the client value, it will never reach the browser
+      | Client => ""
       }
     }
     className={Cx.make([Theme.text(Theme.Color.Gray4)])}>
--- a/packages/react/src/React.mli
+++ b/packages/react/src/React.mli
@@ -517,6 +517,7 @@ module JSX : sig
 
   (** JSX.prop is the representation of HTML/SVG attributes and DOM events *)
   type prop =
+    | Action : (string * string * _ Runtime.server_function) -> prop
     | Bool of (string * string * bool)
     | String of (string * string * string)
     | Style of (string * string * string) list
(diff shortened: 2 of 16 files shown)


demo/universal/native/shared/ServerActionWithFormDataServer.re
```diff
@@ -0,0 +1,23 @@
+[@react.component]
+let make = () => {
+  <form
+    action={
+      switch%platform () {
+      | Server => `Function(ServerFunctions.formData)
+      // doesn't matter the client value, it will never reach the browser
+      | Client => Obj.magic()
```
davesnx: can you make it an empty string?

packages/react/src/React.mli
```diff
@@ -517,6 +517,7 @@ module JSX : sig
 
   (** JSX.prop is the representation of HTML/SVG attributes and DOM events *)
   type prop =
+    | Function : (string * string * _ Runtime.server_function) -> prop
```
davesnx: nit: Can you name it Action?

packages/reactDom/test/test_RSC_html.ml
```diff
@@ -433,6 +433,31 @@ let await_tick ?(raise = false) num =
...
+                  "action",
+                  { Runtime.id = "1234-4321"; call = (fun () -> Lwt.return (React.string "Server Content")) } );
+            ]
+            [ React.string "Server Content" ] )
+  in
+  let main = React.Upper_case_component ("app", app) in
+  assert_html main ~disable_backtrace:true
+    ~shell:
+      "<form>Server Content</form><script data-payload='0:[\"$\",\"form\",null,{\"children\":[\"Server \
+       Content\"],\"action\":\"$F1\"},null,[],{}]\n\
```
davesnx, line 454: I remember discussing this in the past, and now being done is cool. But I don't remember if inlining the action id and action bound works here?

pedrobslisboa: It does not work, react uses the Model Reference ID (`$F`) to handle the function.

https://github.com/facebook/react/blob/a3be6829c6425f306a8bef9f7dba72d1347a64b3/packages/react-client/src/ReactFlightClient.js#L1485-L1495

PR #266, opened 2025-06-25
Demo with no ssr

pedrobslisboa:
## Description

This PR creates a copy of the /router-rsc and makes it work without SSR.

The initial markup comes without the HTML body, but the body is rendered as a model
 - Sample: 
	  ```html
      <html>
        <head>
           <!-- ... -->
        </head>
        <body>
           <!-- Nothing rendered here -->
        </body>
        <script data-payload=[<!-- Body will be available here --> ] />
      </html>
	  ```

## How

At the `to_html` function, we have the' withBodyHtml' flag that removes the body content if it's true. The magic happens on the `Lower_case_element`:

```ocaml
(* ... *)
        if tag = "body" && not withBodyHtml then
          let%lwt _, model = elements_to_html ~fiber children ~withBodyHtml in
          Lwt.return (Html.node tag html_props [], Model.node ~tag ~key ~props:json_props [ model ])
        else if Html.is_self_closing_tag tag then
(* ... *)
```

Some issues that I faced:

Dave pushed this commit [`bad2cdf` (#266)](https://github.com/ml-in-barcelona/server-reason-react/pull/266/commits/bad2cdf84b6a65c79b9dcafabd9fb478648ebcea).
But on running it, I found this error:

image

It happened because we were not providing the HTML wrapper with the initial markup that can be checked in this line: https://github.com/ml-in-barcelona/server-reason-react/commit/bad2cdf84b6a65c79b9dcafabd9fb478648ebcea#diff-bc0d828b259badab876afc8576ac718be0cabea75f26d2d909e2b03306a1d5f9R745

I got the `root_html` back and work directly on the body html instead, as we do not want to have the body html, and we cannot do it at the render_html level. (Do you have another idea?)

https://github.com/ml-in-barcelona/server-reason-react/blob/83eff682326081ff31091bd205933c0370404c18/packages/reactDom/src/ReactServerDOM.ml#L564-L566

```
❌ Error: Can only set one of `children` or `props.dangerouslySetInnerHTML`.
at setProp (chunk-BA4T5M7K.js:12704:25)
at setInitialProperties (chunk-BA4T5M7K.js:13171:98)
at completeWork (chunk-BA4T5M7K.js:8453:28)
at runWithFiberInDEV (chunk-BA4T5M7K.js:1757:72)
at completeUnitOfWork (chunk-BA4T5M7K.js:11248:22)
at performUnitOfWork (chunk-BA4T5M7K.js:11154:29)
at workLoopSync (chunk-BA4T5M7K.js:11000:43)
at renderRootSync (chunk-BA4T5M7K.js:10983:13)
at performWorkOnRoot (chunk-BA4T5M7K.js:10631:46)
at performWorkOnRootViaSchedulerTask (chunk-BA4T5M7K.js:11895:9)

The above error occurred in the <style> component.
React will try to recreate this component tree from scratch using the error boundary you provided, errorBoundary.
```

The solution was not having children when it's null or [null].

```diff
let node ~tag ?(key = None) ~props ?(source = None) ?(debugId = None) ?(owner = None) children : json =
    let key = match key with None -> `Null | Some key -> `String key in
    let debugId = match debugId with None -> `Null | Some debugId -> `String debugId in
    let source = match source with None -> `List [] | Some source -> `List source in
    let owner =
      match owner with
      | None -> `Assoc []
      (* TODO: debugOwner is a ReactComponentInfo itself *)
      | Some owner -> `Assoc [ ("name", `String owner) ]
    in
    let props =
      match children with
      | [] -> props
+     | [ `Null ] -> props
+     | [ `List [ `Null ] ] -> props
      | [ one_children ] -> ("children", one_children) :: props
      | childrens -> ("children", `List childrens) :: props
    in
    `List [ `String "$"; `String tag; key; `Assoc props; debugId; source; owner ]
```


packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -121,6 +121,8 @@ module Model = struct
     let props =
       match children with
       | [] -> props
+      | [ `Null ] -> props
+      | [ `List [ `Null ] ] -> props
```
davesnx: why this is needed?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -690,13 +695,14 @@ let push_children_into html new_children =
...
-let render_html ?(env = `Dev) ?debug:(_ = false) ?bootstrapScriptContent ?bootstrapScripts ?bootstrapModules element =
+let render_html ?(withBodyHtml = true) ?(env = `Dev) ?debug:(_ = false) ?bootstrapScriptContent ?bootstrapScripts
+    ?bootstrapModules element =
   let initial_index = 0 in
   let stream, push, close = Push_stream.make () in
   let context : Fiber.context = { push; close; pending = 1; index = initial_index; env } in
   let is_root_html_node = is_root_html_node element in
   let fiber : Fiber.t = { context; hoisted_head = None; hoisted_head_childrens = []; is_root_html_node } in
-  let%lwt root_html, root_model = to_html ~fiber element in
+  let%lwt root_html, root_model = to_html ~fiber ~withBodyHtml element in
```
davesnx: I preferred the idea of keeping to_html without a flag, did you try that instead?

davesnx: I belive with either a flag or not, this PR needs some unit tests

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: Sorry, I didn't put it on the description

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: > I preferred the idea of keeping to_html without a flag, did you try that instead?

I tried, but the `render_html` wasn't working because we didn't have the root_html:
image

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -654,22 +663,25 @@ let rec to_html ~(fiber : Fiber.t) (element : React.element) : (Html.element * j
...
+let rec is_root_html_node element =
   match (element : React.element) with
-  | Lower_case_element { tag; _ } -> tag = "html"
-  | React.Fragment (React.List [ Lower_case_element { tag = "html"; _ }; _ ]) -> true
-  | _ -> false
+  | Lower_case_element { tag = "html"; _ } -> Lwt.return true
+  | React.Fragment (React.List [ Lower_case_element { tag = "html"; _ }; _ ]) -> Lwt.return true
+  | Async_component (_, component) ->
+      let%lwt element = component () in
+      is_root_html_node element
```
pedrobslisboa: @davesnx turns out we weren't handling the Async html components, so for this line: https://github.com/ml-in-barcelona/server-reason-react/pull/266/files#diff-c42b16a38191961579f8f6e766741e8fee7697090c98a4cec9a890e6fedfcc71R91-R93 we would fall as `not(is_root_html_node)`

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -520,26 +521,34 @@ let rec to_html ~(fiber : Fiber.t) (element : React.element) : (Html.element * j
...
-        let html = Html.node tag html_props [ children ] in
+        let%lwt children_html, children_model = elements_to_html ~fiber children ~withBodyHtml in
+        (* TODO: Could we improve this to avoid the need of normalizing the children model? *)
+        (* elements_to_html always returns a list, and meta/link cannot have children, 
+           if any tag has dangerouslySetInnerHTML, the returned of the moveDangerouslyInnerHtmlAsChildren will be [ InnerHtml innerHtml ] 
+           which is `List [ `Null ] for the model, so we normalize it all `Null *)
+        let normalized_children_model =
+          match children_model with `List [ `Null ] | `List [] -> `Null | _ -> children_model
+        in
+        let html = Html.node tag html_props [ children_html ] in
```
pedrobslisboa: Explaining this:
Without the normalize I was having the following error:
Image
Because the model had children for the meta/link as you can see here:

Image

For the style, the issue was because we had both child and dangerouslySetInnerHTML in the model:
Image

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: @davesnx I could have the withBodyHTML only at the render_html, as you can see now.
Check out those comments: 
https://github.com/ml-in-barcelona/server-reason-react/pull/266#discussion_r2172540396
https://github.com/ml-in-barcelona/server-reason-react/pull/266#discussion_r2172532266

packages/reactDom/src/ReactServerDOM.ml
davesnx: I saw this error happening when there's a problem with the model. Can you try to isolate this issue in another PR or is it related with this?

packages/reactDom/src/ReactServerDOM.ml
davesnx: yeah weird case, we might not want to await for the element in case of async components. Currently an async component being at the root is pretty inusual.

packages/reactDom/src/ReactServerDOM.ml
davesnx: yeah, but I still don't want to populate this flag

pedrobslisboa: Ops, my mistake, the to_html isn't using it anymore, I just forgot to remove

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: Huuum, good point, but the root is async. I'm going to remove it from the root and test it.

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: It was making the demo fail when there was no body html.

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: I changed the RouterRSCNoSSR not to be an async component. But the is_root_html_node was still failing, I identified that we weren't covering this case: https://github.com/ml-in-barcelona/server-reason-react/blob/434574dc1c6d134243988eead38c1dcde63948dd/packages/reactDom/src/ReactServerDOM.ml#L680

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: Just documenting it here.
That's how arch/server produces a simple link tag:
```
["$","link",null,{"rel":"stylesheet","href":"/output.css"},"$1",[["App","/Users/pedrolisboa/work/server-reason-react/arch/server/render-rsc-to-stream.js",58,6]],1]
```
And that is how server-reason-react does
```
["$","link",null,{"children":[],"rel":"stylesheet","href":"/output.css"},null,[],{}]
```

It looks like we shouldn't have any 'children' value (null or []) for elements that don't have it.

PR #270, opened 2025-07-22
Remover InnerHtml and treat dangerouslyInnerHtml separately 

davesnx:
Removes the need for shenanigans with `Null` and keeps logic in the renderer. While this PR adds more noise into render html for Lower_case_elements, it's just because it's a bit more complex than other cases.

### DIFF
--- a/packages/reactDom/test/test_RSC_html.ml
+++ b/packages/reactDom/test/test_RSC_html.ml
@@ -460,11 +460,11 @@ let error_in_toplevel_in_async () =
   let main = React.Async_component ("app", app) in
   assert_raises (Failure "lol") (fun () -> assert_html main ~disable_backtrace:true [])
 
-let await_tick ?(raise = false) num =
+let await_tick ?(raise = false) ?(ms = 10) num =
   React.Async_component
     ( "await_tick",
       fun () ->
-        let%lwt () = sleep ~ms:(Random.int 10) in
+        let%lwt () = sleep ~ms in
         if raise then Lwt.fail (Failure "lol") else Lwt.return (React.string num) )
 
 let server_function_as_action () =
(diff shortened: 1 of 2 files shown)


packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -568,12 +561,13 @@ let rec to_html ~(fiber : Fiber.t) (element : React.element) : (Html.element * j
             attributes
         in
         let json_props = Model.props_to_json json_attributes in
-        if Html.is_self_closing_tag tag then
-          let empty_children = (* there's no children for self closing tags *) [] in
-          Lwt.return (Html.node tag html_props empty_children, Model.node ~tag ~key ~props:json_props empty_children)
-        else
-          let%lwt html, model = elements_to_html ~fiber children in
-          Lwt.return (Html.node tag html_props [ html ], Model.node ~tag ~key ~props:json_props [ model ])
+        match (inner_html, Html.is_self_closing_tag tag) with
```
pedrobslisboa: Running the `arch/server` I found that every tag, not only those on `is_self_closing_tag`, for `<div/>` we have:
```
0:["$","div",null,{},"$1",[],1]
```

Additionally, React throws an error instead of ignoring it as we are doing. For:
```
<meta charSet="utf-8" dangerouslySetInnerHTML={{ __html: "*" }} />
```
We have on:
- SRR: `<meta char-set=\"utf-8\" />`
- React: ``error: meta is a self-closing tag and must neither have `children` nor use `dangerouslySetInnerHTML`.``

I believe we should follow react behavior, WDYT?
Sample:
```ocaml
        match (inner_html, Html.is_self_closing_tag tag) with
        | None, false ->
            if children = [] then Lwt.return (Html.node tag html_props [], Model.node ~tag ~key ~props:json_props [])
            else
              let%lwt html, model = elements_to_html ~fiber children in
              Lwt.return (Html.node tag html_props [ html ], Model.node ~tag ~key ~props:json_props [ model ])
        | Some _, true ->
            raise
              (Invalid_argument
                 (Printf.sprintf
                    "error: %s is a self-closing tag and must neither have `children` nor use `dangerouslySetInnerHTML`.\n"
                    tag))
        | None, true -> Lwt.return (Html.node tag html_props [], Model.node ~tag ~key ~props:json_props [])
        | Some inner_html, false ->
            Lwt.return (Html.node tag html_props [ Html.raw inner_html ], Model.node ~tag ~key ~props:json_props []))
 
```

pedrobslisboa: It could also be nice to have a ppx warning on those self-closing tags, to warning the user.

davesnx: It's already there, breaks the build

pedrobslisboa: Btw, it's there only for children, right? I was mentioning the "dangerouslySetInnerHTML".
I added this line of code to the demo, and it didn't break:
```
    <meta dangerouslySetInnerHTML={"__html": "*"} />
    <meta> {React.string("*")} </meta>
```
Captura de Tela 2025-07-22 às 18 42 50

I don't know why it isn't breaking for the children, though.

PR #271, opened 2025-07-23
Improve `Html.is_self_closing_tag tag` validation

pedrobslisboa:
## Description

This PR improves the `Html.is_self_closing_tag` by adding a condition for `DangerouslyInnerHtml` and improving the message. It also adds tests to it.

### DIFF
--- a/packages/react/src/React.ml
+++ b/packages/react/src/React.ml
@@ -452,8 +452,11 @@ let clone_attributes attributes new_attributes =
 let create_element_with_key ?(key = None) tag attributes children =
   match Html.is_self_closing_tag tag with
   | true when List.length children > 0 ->
-      (* TODO: Add test for this *)
-      raise (Invalid_children "closing tag with children isn't valid")
+      raise (Invalid_children (Printf.sprintf {|"%s" is a self-closing tag and must not have "children".\n|} tag))
+  | true when List.exists (function JSX.DangerouslyInnerHtml _ -> true | _ -> false) attributes ->
+      raise
+        (Invalid_children
+           (Printf.sprintf {|"%s" is a self-closing tag and must not have "dangerouslySetInnerHTML".\n|} tag))
   | true -> Lower_case_element { key; tag; attributes; children = [] }
   | false -> Lower_case_element { key; tag; attributes; children }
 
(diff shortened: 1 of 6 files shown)


demo/universal/native/shared/Counter.re
```diff
@@ -12,6 +12,7 @@ let make = (~initial: int) => {
   };
 
   <Row align=`center gap=2>
+    <meta> {} </meta>
```
davesnx: this seems like an accident?

packages/react/src/React.ml
```diff
@@ -452,8 +452,10 @@ let clone_attributes attributes new_attributes =
 let create_element_with_key ?(key = None) tag attributes children =
   match Html.is_self_closing_tag tag with
   | true when List.length children > 0 ->
-      (* TODO: Add test for this *)
-      raise (Invalid_children "closing tag with children isn't valid")
+      raise (Invalid_children (Printf.sprintf {|"%s" is a self-closing tag and must have "children".\n|} tag))
+  | true when List.exists (function JSX.DangerouslyInnerHtml _ -> true | _ -> false) attributes ->
+      raise
+        (Invalid_children (Printf.sprintf {|"%s" is a self-closing tag and must have "dangerouslySetInnerHTML".\n|} tag))
```
davesnx: and must NOT have

packages/react/src/React.ml
```diff
@@ -452,8 +452,10 @@ let clone_attributes attributes new_attributes =
 let create_element_with_key ?(key = None) tag attributes children =
   match Html.is_self_closing_tag tag with
   | true when List.length children > 0 ->
-      (* TODO: Add test for this *)
-      raise (Invalid_children "closing tag with children isn't valid")
+      raise (Invalid_children (Printf.sprintf {|"%s" is a self-closing tag and must have "children".\n|} tag))
```
davesnx: must NOT have

davesnx (approved):
Comment messages are written wrongly, but the rest looks perfect

demo/universal/native/shared/Counter.re
pedrobslisboa: Indeed an accident.

PR #275, opened 2025-07-25
feat: add React.raw for raw html

pedrobslisboa:
## Description

After removing the React.InnerHtml, we couldn't have raw HTML being rendered on the server

### DIFF
--- a/packages/reactDom/src/ReactServerDOM.ml
+++ b/packages/reactDom/src/ReactServerDOM.ml
@@ -193,6 +193,7 @@ module Model = struct
     let rec turn_element_into_payload ~context element =
       match (element : React.element) with
       | Empty -> `Null
+      | DangerouslyInnerHtml html -> `String html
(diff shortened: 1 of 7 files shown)


packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -428,6 +429,7 @@ let html_suspense_placeholder ~fallback id =
 let rec client_to_html ~fiber (element : React.element) =
   match element with
   | Empty -> Lwt.return Html.null
+  | DangerouslyInnerHtml html -> Lwt.return (Html.raw html)
```
pedrobslisboa, line 432: That's wrong:
Previous:
```
      | InnerHtml _text ->
          raise
            (Invalid_argument
               "InnerHtml does not exist in RSC, this is a bug in server-reason-react.ppx or a wrong construction of \
                JSX manually")
```

PR #278, opened 2025-08-01
Release js packages

pedrobslisboa:
## Description

Release js packages and align server-reason-react pattern


packages/esbuild-plugin/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@ml-in-barcelona/server-reason-react-esbuild-plugin",
   "description": "Esbuild plugin to work with Server Reason React.",
-  "version": "0.0.0",
+  "version": "0.1.0",
```
davesnx, line 4: Do you think we can align this versioning to server-reason-react?

pedrobslisboa: Yep, that's a good idea

Issue #280, opened 2025-08-05
Provide bindings on react-server-dom-esbuild


pedrobslisboa:
## Description

We don't provide any bindings from `react-server-dom-esbuild`, and I think as a Reason project we should have it

PR #283, opened 2025-08-06
Make react-server-dom-esbuild just bindings to .js, abstract callServer

davesnx:
Fixes https://github.com/ml-in-barcelona/server-reason-react/issues/280


demo/universal/js/ReactServerDOMEsbuild.re
```diff
@@ -1,22 +1,21 @@
-type callServer('arg, 'result) =
-  (string, list('arg)) => Js.Promise.t('result);
+type arg;
+type callServer = (string, list(arg)) => Js.Promise.t(React.element);
```
pedrobslisboa, line 2: The `callServer` is also used for server functions that do not return a React element.
This is the callServer type on React: https://github.com/facebook/react/blob/3d14fcf03f4e296d21b52b362f5adefd9e366375/packages/react-client/src/ReactFlightReplyClient.js#L54

demo/universal/js/ReactServerDOMEsbuild.re
pedrobslisboa, line 1: Thanks for the effort on it @davesnx 
You're right about some of the abstractions! To solve the #280 can you move it to the react-server-dom-esbuild and create a dune so we can have it as a melange lib?

PR #285, opened 2025-08-07
Move ReactServerDOMEsbuild.re to packages/

davesnx:


packages/react-server-dom-esbuild/dune
```diff
@@ -0,0 +1,10 @@
+(library
+ (name ReactServerDOMEsbuild)
+ (modules ReactServerDOMEsbuild)
+ (modes melange)
+ (wrapped false)
+ (public_name server-reason-react.react-server-dom-esbuild)
+ (libraries reason-react melange-webapi melange-fetch)
```
pedrobslisboa, line 6:
```suggestion
 (libraries reason-react melange-webapi)
```
Do we need melange-fetch?

packages/react-server-dom-esbuild/dune
```diff
@@ -0,0 +1,10 @@
+(library
+ (name ReactServerDOMEsbuild)
+ (modules ReactServerDOMEsbuild)
+ (modes melange)
+ (wrapped false)
```
pedrobslisboa: As it's not universal, can we make it wrapped?

packages/react-server-dom-esbuild/dune
davesnx: It uses Fetch.response

packages/react-server-dom-esbuild/dune
davesnx: Removed the unwrapped and since we have the same name as the library on the module, the code doesn't change

[`ebe2612` (#285)](https://github.com/ml-in-barcelona/server-reason-react/pull/285/commits/ebe261201c0f1626017c0f782e0c35f8c6734cf2)

PR #286, opened 2025-08-11
Set melange-webapi, melange-fetch and reason-react as dep

pedrobslisboa:


dune-project
```diff
@@ -39,8 +39,12 @@
   (uri (>= 4.2.0))
   (yojson (>= 2.2.0))
   (integers (>= 0.7.0))
+  (melange-fetch (>= 0.1.0))
   (melange-json (>= 2.0.0))
   (melange-json-native (>= 2.0.0))
+  (melange-webapi (>= 0.21.0))
+  (reason-react (>= 0.16.0))
```
davesnx: can you run the build to have an effect on the opam file?

pedrobslisboa: Ops, sorry

PR #293, opened 2025-09-02
feat: add to ppx the root arg

pedrobslisboa:
## Description
This PR tries to re-think the way we create the extract-client-component flag.

## Before

On the extract-client-component flag, we used to change the /js to /native/shared.
But it assumes that native is going to be the source of the shared components, which could not.

## After
By default, the native system replaces /native/shared with nothing, while Melange does the same for /js.
If anyone, for some reason, wants to customize it, it should be done with the --replace_melange_folder_hack and --replace_native_folder_hack args.

### DIFF
--- a/packages/server-reason-react-ppx/server_reason_react_ppx.ml
+++ b/packages/server-reason-react-ppx/server_reason_react_ppx.ml
@@ -6,8 +6,7 @@ type target = Native | Js
 
 (* Since ppxlib doesn't provide a way to get the submodules, we need to keep track of them manually *)
 let mode = ref Native
-let replace_melange_folder_hack = ref "/js"
-let replace_native_folder_hack = ref "/native/shared"
+let shared_folder_prefix = ref "/native/shared"
 let repo_url = "https://github.com/ml-in-barcelona/server-reason-react"
 let issues_url = Printf.sprintf "%s/issues" repo_url
 
@@ -733,7 +732,7 @@ module ServerFunction = struct
 
   let generate_id ~loc name =
     (* We need to add a nasty hack here, since have different files for native and melange.Assume that the file structure is native/lib and js, and replace the name directly. This is supposed to be temporal, until dune implements https://github.com/ocaml/dune/issues/10630 *)
-    let file_path = loc.loc_start.pos_fname |> Str.replace_first (Str.regexp replace_native_folder_hack.contents) "" in
+    let file_path = loc.loc_start.pos_fname |> Str.replace_first (Str.regexp shared_folder_prefix.contents) "" in
     let hash = Printf.sprintf "%s_%s_%d" name file_path loc.loc_start.pos_lnum |> Hashtbl.hash |> string_of_int in
     hash
 
@@ -956,7 +955,7 @@ let rewrite_structure_item ~nested_module_names structure_item =
               let loc = expr.pexp_loc in
               let file =
                 expr.pexp_loc.loc_start.pos_fname
-                |> Str.replace_first (Str.regexp replace_native_folder_hack.contents) ""
+                |> Str.replace_first (Str.regexp shared_folder_prefix.contents) ""
                 |> estring ~loc
               in
               let import_module =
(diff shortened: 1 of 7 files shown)


packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -952,10 +954,14 @@ let rewrite_structure_item ~nested_module_names structure_item =
         if isReactClientComponentBinding vb then
           expand_make_binding vb (fun expr ->
               let loc = expr.pexp_loc in
-              let file = pexp_ident ~loc { txt = Lident "__FILE__"; loc } in
+              let file =
```
pedrobslisboa, line 956: @davesnx I change it from `__FILE__` to get the file from loc with `pos_fname`. Is it ok?

davesnx: Sure, but why?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -1102,5 +1108,9 @@ let traverse =
 
 let () =
   Driver.add_arg "-melange" (Unit (fun () -> mode := Js)) ~doc:"preprocess for js build";
+  Driver.add_arg "--replace-melange-folder-hack"
+    (String (fun str -> replace_melange_folder_hack := str))
+    ~doc:"root path";
```
davesnx: can we have a decent description?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: To have the replace on ppx time instead of runtime.

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: Sure, my bad

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -1102,5 +1108,9 @@ let traverse =
 
 let () =
   Driver.add_arg "-melange" (Unit (fun () -> mode := Js)) ~doc:"preprocess for js build";
+  Driver.add_arg "--replace-melange-folder-hack"
```
davesnx: wdut about --copyfiles-melange-folder-prefix?

pedrobslisboa: ~Looks better to me.~

Maybe
--shared-files-melange-folder-prefix
--shared-files-native-folder-prefix

packages/server-reason-react-ppx/server_reason_react_ppx.ml
davesnx: Oh, we need the replace here as well? 

There's a funny thing here that if 2 components have same name but one in js and one in native, weird things can happen if we replace both.

packages/server-reason-react-ppx/server_reason_react_ppx.ml
davesnx: yeah, but files and folder in the same name is bad and also, shared doesn't make much sense here.

```
--melange-folder-prefix
--native-folder-prefix 
```

wdyt?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: You mean two modules file, one `js` only and other `native` only, with the same name, and also as both client components?
If someone does something like this I think they are with a strange mindset, too specific and looks wrong.
Why someone would like to have one `js` only client component and another `native` only client component that aren't the same, but with the same name?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: --melange-folder-prefix and --native-folder-prefix I think don't fit, for example, we can have `/native/shared/Button.re` and copy it to js having it as `/js/Button.re`. In the end, we would need to replace `/native/shared` by "" and `/js` by "". And `/native/shared` isn't the folder prefix on native and can lead to a misunderstanding of the flag.

pedrobslisboa: Maybe: --universal-melange-folder-prefix (Where is the folder that will contain the universal code?)
Or: --shared-melange-folder-prefix (Where is the folder that will contain the shared code?)

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -6,6 +6,7 @@ type target = Native | Js
 
 (* Since ppxlib doesn't provide a way to get the submodules, we need to keep track of them manually *)
 let mode = ref Native
+let shared_folder_prefix = ref "/native/shared"
```
pedrobslisboa, line 9: @davesnx I turned it into a unified arg.

davesnx (changes requested):
I don't see a change in the demo, so I assume it will break there.

PR #294, opened 2025-09-04
feat: add to ppx the root arg

davesnx:
This is the resurrected PR https://github.com/ml-in-barcelona/server-reason-react/pull/293


packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -731,13 +737,16 @@ module ServerFunction = struct
...
     let replacement =
       match shared_folder_prefix.contents with
-      | Some x -> x
+      | Some x ->
+          if match_substring file_path x then x
+          else failwith "Prefix doesn't match the file path. Provide a prefix that matches the file path."
       | None -> failwith "Found a server.function without --shared-folder-prefix argument. Provide one."
     in
+    let file_path = Str.replace_first (Str.regexp replacement) "" file_path in
     (* We need to add a nasty hack here, since have different files for native and melange.Assume that the file structure is native/lib and js, and replace the name directly. This is supposed to be temporal, until dune implements https://github.com/ocaml/dune/issues/10630 *)
```
davesnx, line 748: can you leave the comment in the right place?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -731,13 +737,16 @@ module ServerFunction = struct
     | _ -> fn
 
   let generate_id ~loc name =
+    let file_path = loc.loc_start.pos_fname in
     let replacement =
       match shared_folder_prefix.contents with
-      | Some x -> x
+      | Some x ->
+          if match_substring file_path x then x
+          else failwith "Prefix doesn't match the file path. Provide a prefix that matches the file path."
```
davesnx: wdyt about not failing with, and have a location error?

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -10,6 +10,12 @@ let shared_folder_prefix = ref None
 let repo_url = "https://github.com/ml-in-barcelona/server-reason-react"
 let issues_url = Printf.sprintf "%s/issues" repo_url
 
+let match_substring string substring =
+  try
+    Str.search_forward (Str.regexp substring) string 0 |> ignore;
```
davesnx: Can you use Str.regexp_string instead of regexp to not have any surprising issues?

pedrobslisboa: Sure

packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -1100,7 +1131,17 @@ let traverse =
           with Error err -> [%expr [%e err]])
   end
 
+let routes = Routes.Parts.of_parts "/routes"
```
davesnx: This shouldn't be here I guess

packages/server-reason-react-ppx/server_reason_react_ppx.ml
pedrobslisboa: As it returns a string (we need that string further), and to have a location error we would need an expression I kept the fail and treated this error on calling the function, take a look: [`249712a` (#294)](https://github.com/ml-in-barcelona/server-reason-react/pull/294/commits/249712ae1fa5eb90d3d0227aec619dcf34a6e53a)

davesnx: need to use raise_errorf

PR #297, opened 2025-09-04
Change ppx execution order

davesnx:
Run styled expansion in server-reason-react


packages/server-reason-react-ppx/server_reason_react_ppx.ml
```diff
@@ -1145,5 +1145,6 @@ let () =
          let prefix = if prefix = "" then "" else prefix ^ "/" in
          shared_folder_prefix := Some prefix))
     ~doc:"prefix of shared folder, used to replace the it in the file path";
-  Ppxlib.Driver.V2.register_transformation "server-reason-react.ppx" ~preprocess_impl:traverse#structure
+  Driver.V2.register_transformation "server-reason-react.ppx"
+    ~instrument:(Driver.Instrument.V2.make ~position:Before (fun structure -> traverse#structure structure))
```
pedrobslisboa: Why is that a need?
Also, why only on the structure and not on the signature?

PR #300, opened 2025-09-16
Fix/html resources

pedrobslisboa:
## Description

This PR improves the way we create our HTML for RSC, it follows the output we can simulate on `arch/server`


packages/html/Html.ml
```diff
@@ -70,7 +70,32 @@ type element =
...
+      | [] -> None
+      | Node node :: rest ->
+          let next_level = cur_level +. 1. in
+          if node.tag = tag then Some node
+          else aux tag next_level (if next_level > level then rest else node.children @ rest)
+      | List (_, children) :: rest -> aux tag cur_level (children @ rest)
+      | Array children :: rest -> aux tag cur_level (Array.to_list children @ rest)
+      | _ :: rest -> aux tag cur_level rest
+    in
+    aux tag 0. [ tree ]
```
davesnx, line 91: The find_by_tag fn is very similar on the approach of traversing the three again, but running 2 times?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -662,75 +653,48 @@ and render_lower_case_element ~fiber ~key ~tag ~attributes ~children =
...
-    if (* disable_model *) true then
-      (* Currently we don't sent the model for those cases, since we hydrate the document.body as soon as we hydrate the entire document *)
-      `Null
-    else
-      (* In case of the model, we don't care about inner_html as a children since we need it as a prop. This is the opposite from html rendering *)
-      match (Html.is_self_closing_tag tag, inner_html) with
-      | _, Some _ | true, _ -> Model.node ~tag ~key ~props []
-      | false, None -> Model.node ~tag ~key ~props [ children ]
+  (* Unlike react, we don't reorder the elements in the head, we keep the original order to avoid hydration errors *)
+  (* TODO: Reorder the elements in the head to fit react's behavior? *)
```
davesnx, line 658: afaik React shouldn't complain for hydration errors on the head. And also, we should "order" them in the way priorities go

packages/html/test/test_node.ml
```diff
@@ -0,0 +1,46 @@
+open Html
```
davesnx, line 1: what's this test about?

packages/html/Html.ml
```diff
@@ -70,7 +70,32 @@ type element =
   | List of (string * element list)
   | Array of element array
 
-and node = { tag : string; attributes : attribute_list; children : element list }
+and node = { tag : string; attributes : attribute_list; mutable children : element list }
```
davesnx, line 73: I don't think a mutable children is what we want.

packages/html/Html.ml
pedrobslisboa: We don't traverse the entire tree again as the find_by_tag stops at the first tag is found.

packages/html/test/test_node.ml
pedrobslisboa: I test the `Html.Node`.

PR #305, opened 2025-09-18
Fix mismatch of the model and html on render_html

pedrobslisboa:
## Description

This PR keeps the way we reconstruct the html, but changes the model construct.

## Notes

-  We are now collecting the `<html>` attributes as we are rebuilding it.
- All resources now are `Html.node` instead of a custom type, which makes it simpler to maintain and is compatible with the common type we have across the project.
- The head html now is also pushed as a `Html.node option` instead of the custom type `mutable hoisted_head : (Html.attribute_list * Html.element list) option;`
- To keep the model as declared, instead of handling only the `<html>` children with `render_element_to_html ~fiber (React.List children)`, we use `render_regular_element` to render it, but only pass it to the `<html>` only the children:
 ```ocaml
      | Some "html" ->
          Fiber.set_html_tag_attributes ~fiber (ReactDOM.attributes_to_html attributes);
          let%lwt html, model = render_regular_element ~fiber ~key ~tag ~attributes ~children ~inner_html in
          let html_children = match html with Html.Node { children; _ } -> Html.list children | _ -> html in
          Lwt.return (html_children, model)
  ```
- On the `create_model` we don't render `` `Null `` anymore.

### DIFF
--- a/packages/reactDom/src/ReactServerDOM.ml
+++ b/packages/reactDom/src/ReactServerDOM.ml
@@ -696,19 +696,19 @@ and render_lower_case_element ~fiber ~key ~tag ~attributes ~children =
         fiber.inside_head <- true;
         (* push the head element to the hoisted_head *)
         let%lwt value =
-          render_hoisted_element ~fiber ~key ~tag ~attributes ~children ~inner_html ~on_push:Fiber.push_hoisted_head
+          handle_hoistable_element ~fiber ~key ~tag ~attributes ~children ~inner_html ~on_push:Fiber.push_hoisted_head
         in
         fiber.inside_head <- false;
         Lwt.return value
     | tag when (tag = "script" && is_async attributes) || (tag = "link" && has_precedence_and_rel_stylesheet attributes)
       ->
-        render_hoisted_element ~fiber ~key ~tag ~attributes ~children ~inner_html ~on_push:Fiber.push_resource
+        handle_hoistable_element ~fiber ~key ~tag ~attributes ~children ~inner_html ~on_push:Fiber.push_resource
     | tag when tag = "title" || tag = "meta" || tag = "link" ->
-        render_hoisted_element ~fiber ~key ~tag ~attributes ~children ~inner_html
+        handle_hoistable_element ~fiber ~key ~tag ~attributes ~children ~inner_html
           ~on_push:Fiber.push_hoisted_head_childrens
     | _ -> render_regular_element ~fiber ~key ~tag ~attributes ~children ~inner_html
 
-and render_hoisted_element ~fiber ~key ~tag ~attributes ~children ~inner_html ~on_push =
+and handle_hoistable_element ~fiber ~key ~tag ~attributes ~children ~inner_html ~on_push =
   let props = Model.props_to_json attributes in
 
   let create_model children =
--- a/packages/reactDom/src/ReactServerDOM.ml
+++ b/packages/reactDom/src/ReactServerDOM.ml
@@ -23,27 +23,24 @@ let create_stack_trace () =
 
 (* Resources module maintains insertion order while deduplicating based on src/href *)
 module Resources = struct
-  type item = string * Html.attribute_list * Html.element option
-  type t = item list
-
   let get_attribute ~key:key_to_get (attributes : Html.attribute_list) =
     List.find_map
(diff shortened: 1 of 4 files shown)


packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -60,20 +57,25 @@ module Fiber = struct
     (* visited_first_lower_case stores the tag of the first lower case element visited, useful to know if the root element is an html tag *)
     mutable visited_first_lower_case : string option;
     (* hoisted_head stores the <head> element's attributes and direct children *)
-    mutable hoisted_head : (Html.attribute_list * Html.element list) option;
+    mutable hoisted_head : Html.node option;
```
davesnx, line 60: If I undertsand it correctly "hoisted_head" is now the entire node? 

Why do we keep `html_tag_attributes` as a separate field then?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -60,20 +57,25 @@ module Fiber = struct
...
   }
 
-  let push_hoisted_head ~fiber html_attributes children = fiber.hoisted_head <- Some (html_attributes, children)
+  let set_html_tag_attributes ~fiber html_attributes = fiber.html_tag_attributes <- html_attributes
+  let push_hoisted_head ~fiber head = fiber.hoisted_head <- Some head
   let push_resource ~fiber resource = fiber.resources <- Resources.add resource fiber.resources
 
   let push_hoisted_head_childrens ~fiber children =
-    fiber.hoisted_head_childrens <- fiber.hoisted_head_childrens @ [ children ]
+    fiber.hoisted_head_childrens <- fiber.hoisted_head_childrens @ [ Node children ]
```
davesnx, line 78: Do we still use the Node constructor?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -662,75 +664,71 @@ and render_lower_case_element ~fiber ~key ~tag ~attributes ~children =
...
+        Lwt.return value
+    | tag when (tag = "script" && is_async attributes) || (tag = "link" && has_precedence_and_rel_stylesheet attributes)
+      ->
+        render_hoisted_element ~fiber ~key ~tag ~attributes ~children ~inner_html ~on_push:Fiber.push_resource
+    | tag when tag = "title" || tag = "meta" || tag = "link" ->
+        render_hoisted_element ~fiber ~key ~tag ~attributes ~children ~inner_html
+          ~on_push:Fiber.push_hoisted_head_childrens
+    | _ -> render_regular_element ~fiber ~key ~tag ~attributes ~children ~inner_html
+
+and render_hoisted_element ~fiber ~key ~tag ~attributes ~children ~inner_html ~on_push =
```
davesnx: instead of render, what do you think of "hoist_element" and the "on_push" -> "into"

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: `html_tag_attributes` is for the the html. We need to collect those attributes as we rebuild the HTML tag.

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: Huuum, but the `render_regular_element` also returns a value, which is `Lwt.return (Html.null, create_model_node children_model`, it doesn't only hoist

davesnx: I mean, it does not render an html, that's for sure xd

pedrobslisboa: > I mean, it does not render an html, that's for sure xd

True, but it still renders the model.

Maybe `handle_hoistable_element` instead?

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: Yep, because the children of a Node are `Html.element`s, as we got the node we need to push it as an element.

packages/reactDom/src/ReactServerDOM.ml
davesnx: it's very nick pick, but yea hoist_element is good enough

PR #306, opened 2025-09-29
Make client component to execute in runtime

pedrobslisboa:
## Description

On studying and investigating how to apply an RSC navigation, I fell into an issue, where I wasn't able to render nested Providers and Consumers of the same Context.

## Problem

Take a look at that code:

```js
// Client.js
'use client'

import React, { createContext, useContext } from 'react'

const Context = createContext(<div>Hello</div>)

export const Provider = ({ value, children }) => {
  return <Context.Provider value={value}>{children}</Context.Provider>
}

export const Outlet = () => {
  const context = useContext(Context)
  return context
}

// App.js
const Content = () => {
  return (
      /content
  )
}

const About = () => {
  return (
    <Provider value={<Content />}>
      /about
      <Outlet />
    </Provider>
  )
}

const Root = () => {
  return (
    <Provider value={<About />}>
      Root
      <Outlet />
    </Provider>
  )
}

const App = () => {
  <div>
    <Root />
  </div>
}

ReactDOM.renderToReadableStream(<App />).then((stream) => {
	debug(stream);
});
```

The result of it will be: `<div>Root<!-- -->/about<!-- -->/content</div>`

However, when I had something similar on `server-reason-react`, the result was `<div>Root</div>`. That's because the client component wasn't running at the time of the request, but as soon as the `Client_component` is created, the  <Outlet /> represents `React.null` as the initial value, without waiting for the Provider value.

## Solution

To make it work properly, the client value on `Client_component` is now lazy.
```
  | Client_component of { props : client_props; client : unit -> element; import_module : string; import_name : string }
```

### DIFF
--- a/packages/reactDom/test/test_RSC_html.ml
+++ b/packages/reactDom/test/test_RSC_html.ml
@@ -225,7 +225,7 @@ let suspenasync_and_client () =
                       React.Client_component
                         {
                           props = [];
-                          client = React.string "Only the client";
+                          client = (fun () -> React.string "Only the client");
                           import_module = "./client-with-props.js";
                           import_name = "";
                         };
@@ -301,7 +301,7 @@ let client_with_promise_props () =
                 {
                   props =
                     [ ("promise", React.Promise (delayed_value ~ms:20 "||| Resolved |||", fun res -> `String res)) ];
-                  client = React.string "Client with Props";
+                  client = (fun () -> React.string "Client with Props");
                   import_module = "./client-with-props.js";
                   import_name = "ClientWithProps";
                 };
@@ -333,7 +333,7 @@ let client_with_element_props () =
                       (React.createElement "span" [] [ React.string "server-component-as-props-to-client-component" ])
                   );
                 ];
-              client = React.string "Client with elment prop";
+              client = (fun () -> React.string "Client with elment prop");
               import_module = "./client-with-props.js";
               import_name = "ClientWithProps";
             } )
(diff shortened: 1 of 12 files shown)


packages/reactDom/test/test_RSC_html.ml
```diff
@@ -239,8 +239,8 @@ let suspenasync_and_client () =
        data-payload='0:[\"$\",\"$Sreact.suspense\",null,{\"fallback\":\"Loading...\",\"children\":\"$L1\"},null,[],{}]\n\
        '>window.srr_stream.push()</script>"
     [
-      "<script data-payload='2:I[\"./client-with-props.js\",[],\"\"]\n'>window.srr_stream.push()</script>";
-      "<div hidden=\"true\" id=\"S:1\"><span>Only the client<!-- -->Part of async component</span></div>\n\
+      "<script data-payload='2:I[\"./provider-with-props.js\",[],\"\"]\n'>window.srr_stream.push()</script>";
+      "<div hidden=\"true\" id=\"S:1\"><span>Only the provider<!-- -->Part of async component</span></div>\n\
```
davesnx: can you keep the names as before?

davesnx (approved):
Make sense!

packages/reactDom/test/test_RSC_html.ml
pedrobslisboa: Ops, my mistake on find-n-replace

PR #312, opened 2025-10-11
Improve the Fiber and Model stream context

pedrobslisboa:
## Description

This PR refactors the way we handle the context for Model and Fiber. Instead of 2 different contexts for the Stream, we will have the same context type for `both`.
```ocaml
type 'a context = {
  push : 'a -> unit;
  close : unit -> unit;
  mutable index : int;
  mutable pending : int;
  env : env;
  debug : bool;
  }
```

The context controls async and sync push as the index and pending, so we no longer need to worry about it. We declare the value to push and use `push/push_async`, and the context will handle the rest:

```ocaml
  let push chunk ~context =
    let index = context.index in
    context.push (chunk index);
    context.index <- context.index + 1;
    index

  let push_async promise ~context =
    let index = context.index in
    context.index <- context.index + 1;
    context.pending <- context.pending + 1;
    Lwt.async (fun () ->
        let%lwt chunk = promise in
        context.pending <- context.pending - 1;
        context.push (chunk index);
        if context.pending = 0 then context.close ();
        Lwt.return ());
    index
```

The concept of inversion of control was applied at `render_element_to_html and `client_values_to_json`. Then we can use the `client_values_to_json` at `render_element_to_html, which reduces the amount of code to maintain, also keeping it way simpler to understand:

<table>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
<tr>
<td>

```ocaml
| Client_component { import_module; import_name; props; client } ->
    let lwt_props =
      Lwt_list.map_p
        (fun (name, value) ->
          match (value : React.client_value) with
          | Element element ->
              let%lwt _html, model = render_element_to_html ~fiber element in
              Lwt.return (name, model)
          | Promise (promise, value_to_json) ->
                let%lwt value = promise in
                let json = value_to_json value in
                let ret = chunk_script (Model.model_to_chunk index json) in
                Lwt.return ret
              in
              context.pending <- context.pending + 1;
              Lwt.async (fun () ->
                  let%lwt html = async in
                  context.push html;
                  context.pending <- context.pending - 1;
                  if context.pending = 0 then context.close ();
                  Lwt.return ());
              Lwt.return (name, `String (Model.promise_value index))
          | Json json -> Lwt.return (name, json)
          | Error error ->
              let error_json =
                Model.make_error_json 
                     ~env:context.env 
                     ~stack:error.stack 
                     ~message:error.message 
                     ~digest:error.digest
              in
              context.push (error_chunk_script index error_json);
              Lwt.return (name, `String (Model.error_value index))
          | Function action ->
              let html =
                chunk_script 
                         (Model.model_to_chunk index 
                         (`Assoc [ ("id", `String action.id); ("bound", `Null) ]))
              in
              context.push html;
              Lwt.return (name, `String (Model.action_value index)))
        props
    in
    let lwt_html = client_to_html ~fiber (client ()) in

    let ref : json = Model.component_ref 
                                 ~module_:import_module 
                                 ~name:import_name in
    context.push (client_reference_chunk_script index ref);
    let%lwt html, props = Lwt.both lwt_html lwt_props in
    let model = Model.node ~tag:(Model.ref_value index) ~props [] in
    Lwt.return (html, model)
```

</td>
<td>

```ocaml
| Client_component { import_module; import_name; props; client } ->
    let context = get_context fiber in
    let props =
      let on_push value = on_push (Model value) in
      Model.client_values_to_json ~context:fiber.context ~on_push props
    in
    let%lwt html = client_to_html ~fiber ~on_push (client ()) in
    let ref : json = Model.component_ref 
                                 ~module_:import_module 
                                 ~name:import_name in
    let index = Stream.push ~context (on_push (Model (Component_ref ref))) in
    let model = Model.node ~tag:(Model.ref_value index) ~props [] in
    Lwt.return (html, model)		
```

</td>
</tr>
</table>

### DIFF
--- a/packages/reactDom/src/ReactServerDOM.ml
+++ b/packages/reactDom/src/ReactServerDOM.ml
@@ -248,7 +248,7 @@ module Model = struct
   let rec element_to_payload ~context ?(debug = false) ~to_chunk ~env element =
     let is_root = ref true in
 
-    let rec turn_element_into_payload element =
+    let rec turn_element_into_payload ~context element =
       match (element : React.element) with
       | Empty -> `Null
       | DangerouslyInnerHtml _ ->
@@ -272,18 +272,18 @@ module Model = struct
               attributes
           in
           let props = props_to_json attributes in
-          node ~key ~tag ~props (List.map turn_element_into_payload children)
+          node ~key ~tag ~props (List.map (turn_element_into_payload ~context) children)
       | Fragment children ->
           if is_root.contents then is_root := false;
-          turn_element_into_payload children
+          turn_element_into_payload ~context children
       | List children ->
           if is_root.contents then is_root := false;
-          `List (List.map turn_element_into_payload children)
+          `List (List.map (turn_element_into_payload ~context) children)
       | Array children ->
           if is_root.contents then is_root := false;
-          `List (Array.map turn_element_into_payload children |> Array.to_list)
-      (* TODO: Get the stack info from component as well *)
+          `List (Array.map (turn_element_into_payload ~context) children |> Array.to_list)
       | Upper_case_component (name, component) -> (
+          (* TODO: Get the stack info from component *)
           match component () with
           | element ->
               (* TODO: Can we remove the is_root difference. It currently align with react.js behavior, but it's not clear what is the purpose of it *)
@@ -294,13 +294,13 @@ module Model = struct
                   Root is a special case: https://github.com/facebook/react/blob/f3a803617ec4ba9d14bf5205ffece28ed1496a1d/packages/react-server/src/ReactFlightServer.js#L756-L766 
                 *)
                 if debug then push_debug_info ~context ~to_chunk ~env ~index:0 ~ownerName:name else ();
-                turn_element_into_payload element)
+                turn_element_into_payload ~context element)
               else
                 (* If it's not the root React push the element to the stream and return the reference value *)
                 let element_index =
                   Stream.push ~context (fun index ->
                       if debug then push_debug_info ~context ~to_chunk ~env ~index ~ownerName:name else ();
-                      let payload = turn_element_into_payload element in
+                      let payload = turn_element_into_payload ~context element in
                       to_chunk (Value payload) index)
                 in
                 `String (ref_value element_index)
@@ -318,12 +318,12 @@ module Model = struct
               let error = make_error ~message ~stack ~digest:"" in
               let index = Stream.push ~context (to_chunk (Error (env, error))) in
               `String (lazy_value index)
-          | Return element -> turn_element_into_payload element
+          | Return element -> turn_element_into_payload ~context element
           | Sleep ->
               let promise =
                 try%lwt
                   let%lwt element = promise in
-                  Lwt.return (to_chunk (Value (turn_element_into_payload element)))
+                  Lwt.return (to_chunk (Value (turn_element_into_payload ~context element)))
                 with exn ->
                   let message = Printexc.to_string exn in
                   let stack = create_stack_trace () in
@@ -335,18 +335,18 @@ module Model = struct
       | Suspense { key; children; fallback } ->
           (* TODO: Need to check is_root? *)
           (* TODO: Maybe we need to push suspense index and suspense node separately *)
-          let fallback = turn_element_into_payload fallback in
-          suspense_node ~key ~fallback [ turn_element_into_payload children ]
+          let fallback = turn_element_into_payload ~context fallback in
+          suspense_node ~key ~fallback [ turn_element_into_payload ~context children ]
       | Client_component { import_module; import_name; props; client = _ } ->
           let ref = component_ref ~module_:import_module ~name:import_name in
           let index = Stream.push ~context (to_chunk (Component_ref ref)) in
           let client_props = client_values_to_json ~context ~to_chunk ~env props in
           node ~tag:(ref_value index) ~props:client_props []
       (* TODO: Do we need to do anything with Provider and Consumer? *)
-      | Provider children -> turn_element_into_payload children
-      | Consumer children -> turn_element_into_payload children
+      | Provider children -> turn_element_into_payload ~context children
+      | Consumer children -> turn_element_into_payload ~context children
     in
-    turn_element_into_payload element
+    turn_element_into_payload ~context element
 
   and client_value_to_json ~context ?debug ~to_chunk ~env value =
     match (value : React.client_value) with
@@ -355,9 +355,8 @@ module Model = struct
         let index = Stream.push ~context (to_chunk (Error (env, error))) in
         `String (error_value index)
     | Element element ->
-        let index =
-          Stream.push ~context (to_chunk (Value (element_to_payload ~context ?debug ~to_chunk ~env element)))
-        in
+        let payload = element_to_payload ~context ?debug ~to_chunk ~env element in
+        let index = Stream.push ~context (to_chunk (Value payload)) in
         `String (ref_value index)
     | Promise (promise, value_to_json) -> (
         match Lwt.state promise with
--- a/packages/reactDom/test/test_RSC_model.ml
+++ b/packages/reactDom/test/test_RSC_model.ml
@@ -714,16 +714,17 @@ let env_development_adds_debug_info () =
     React.Upper_case_component
       ( "app",
         fun () ->
+          let value = "my friend" in
           React.createElement "input"
             [
               React.JSX.String ("id", "id", "sidebar-search-input");
               React.JSX.String ("placeholder", "placeholder", "Search");
-              React.JSX.String ("value", "value", "my friend");
+              React.JSX.String ("value", "value", value);
             ]
             [] )
   in
   let output, subscribe = capture_stream () in
-  let%lwt () = ReactServerDOM.render_model ~debug:true ~subscribe app in
+  let%lwt () = ReactServerDOM.render_model ~subscribe ~debug:true app in
   assert_list_of_strings !output
     [
       "1:{\"name\":\"app\",\"env\":\"Server\",\"key\":null,\"owner\":null,\"stack\":[],\"props\":{}}\n";


packages/reactDom/test/test_RSC_model.ml
```diff
@@ -147,9 +147,7 @@ let upper_case_with_list () =
   let%lwt () = ReactServerDOM.render_model ~subscribe (app ()) in
   assert_list_of_strings !output
     [
-      "1:[\"$\",\"span\",null,{\"children\":\"hi\"},null,[],{}]\n";
-      "2:[\"$\",\"span\",null,{\"children\":\"hola\"},null,[],{}]\n";
-      "0:[\"$1\",\"$2\"]\n";
```
davesnx, line 152: are these changes "on purpose"?

packages/reactDom/test/test_RSC_model.ml
```diff
@@ -725,7 +718,7 @@ let env_development_adds_debug_info () =
   assert_list_of_strings !output
     [
       "1:{\"name\":\"app\",\"env\":\"Server\",\"key\":null,\"owner\":null,\"stack\":[],\"props\":{}}\n";
-      "0:D\"$1\"\n";
+      "2:D\"$1\"\n";
       "0:[\"$\",\"input\",null,{\"id\":\"sidebar-search-input\",\"placeholder\":\"Search\",\"value\":\"my \
```
davesnx: whoa, this was a bug?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -21,6 +22,39 @@ let create_stack_trace () =
...
+    let index = context.index in
+    context.index <- context.index + 1;
+    context.pending <- context.pending + 1;
+    Lwt.async (fun () ->
+        let%lwt chunk = promise in
+        context.pending <- context.pending - 1;
+        context.push (chunk index);
+        if context.pending = 0 then context.close ();
+        Lwt.return ());
+    index
```
davesnx: Looks like a decent abstraction, but caused some tests to behave differently. (since we pushed a few things instead of push once)

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -43,67 +77,9 @@ module Resources = struct
...
-  let use_index t =
-    t.context.index <- t.context.index + 1;
-    t.context.index
-
-  let get_context t = t.context
-end
-
 module Model = struct
-  type chunk_type = Chunk_value of json | Chunk_component_ref of json | Debug_info_map of json | Chunk_error of json
+  type t = Value of json | Debug_info_map of json | Component_ref of json | Error of env * React.error
```
davesnx: Chunk_value and Chunk_component_ref had different serialisation right? How come we join them?

davesnx: Never mind the diff confused me. Chunk_value -> Value looks... interesting?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,178 +226,430 @@ module Model = struct
               (fun prop ->
                 match prop with
                 | React.JSX.Action (_, key, f) ->
-                    let id = use_chunk_id context in
-                    context.push id (Chunk_value (`Assoc [ ("id", `String f.id); ("bound", `Null) ]));
-                    React.JSX.String (key, key, action_value id)
+                    let index =
+                      Stream.push ~context (on_push (Value (`Assoc [ ("id", `String f.id); ("bound", `Null) ])))
```
davesnx: Stream.push ~context onpush?

shouldn't be just "on_push"?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,178 +226,430 @@ module Model = struct
...
-                    let message = Printexc.to_string exn in
-                    let stack = create_stack_trace () in
-                    let error_json = make_error_json ~env:context.env ~message ~stack ~digest:"" in
-                    context.push index (Chunk_error error_json);
-                    context.pending <- context.pending - 1;
-                    Lwt.return ());
+              let promise =
+                try%lwt
+                  let%lwt element = promise in
+                  Lwt.return (on_push (Value (turn_element_into_payload element)))
```
davesnx: What, on_push returns a unit right? Could you split into a separate line and keep the Lwt.return ()?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,178 +226,430 @@ module Model = struct
...
+                with exn ->
+                  let message = Printexc.to_string exn in
+                  let stack = create_stack_trace () in
+                  let error = make_error ~message ~stack ~digest:"" in
+                  Lwt.return (on_push (Error (context.env, error)))
+              in
+              let index = Stream.push_async promise ~context in
```
davesnx, line 333: Def not a big fan of this abstraction, I would consider keeping it as it is, wdyt? (Using the onpush, I mean)

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,178 +226,430 @@ module Model = struct
...
-        `String (ref_value chunk_id)
+        let index = Stream.push ~context (on_push (Error (context.Stream.env, error))) in
+        `String (error_value index)
+    | Element element -> element_to_payload ~context ~on_push element
     | Promise (promise, value_to_json) -> (
         match Lwt.state promise with
-        | Return value ->
-            let chunk_id = use_chunk_id context in
-            let json = value_to_json value in
-            context.push context.chunk_id (Chunk_value json);
```
davesnx, line 351: This doesn't push anymore?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,178 +226,430 @@ module Model = struct
...
-    let initial_chunk_id = 0 in
-    let stream, push, close = Push_stream.make () in
-    let push_chunk id chunk =
-      match chunk with
-      | Chunk_value json -> push (model_to_chunk id json)
-      | Debug_info_map json -> push (debug_info_to_chunk id json)
-      | Chunk_component_ref json -> push (client_reference_to_chunk id json)
-      | Chunk_error json -> push (error_to_chunk id json)
+module Fiber = struct
+  type chunk = Model of Model.t | Suspense_Boundary of Html.element | Batch of chunk list
```
davesnx: how's possible that chunk contains Suspense_boundary (boundary could be lowercase) but context can stream Html.element? 

This looks like an extra diff from the refactor?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,178 +226,430 @@ module Model = struct
...
+      [
+        Html.node "div" [ Html.attribute "hidden" "true"; Html.attribute "id" (Printf.sprintf "S:%x" index) ] [ html ];
+        rc_replacement index index;
+      ]
+
+  let rec to_chunk chunk id =
+    match chunk with
+    | Model model -> model_to_chunk model id
+    | Suspense_Boundary str -> boundary_to_chunk id str
+    | Batch chunks -> chunks |> List.map (fun chunk -> to_chunk chunk id) |> Html.list
```
davesnx: When Batch is needed?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -839,13 +746,10 @@ let render_html ?(skipRoot = false) ?(env = `Dev) ?debug:(_ = false) ?bootstrapS
       inside_body = false;
     }
   in
-  let%lwt root_html, root_model = render_element_to_html ~fiber element in
-  let root_chunk = client_value_chunk_script initial_index root_model in
-  context.pending <- context.pending - 1;
+  let%lwt root_html, root_model = Fiber.render_element_to_html ~fiber ~on_push:Fiber.to_chunk element in
```
davesnx: If we send fiber, why send on_push as a labelled arg?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -825,8 +733,7 @@ let render_html ?(skipRoot = false) ?(env = `Dev) ?debug:(_ = false) ?bootstrapS
               modules
         | None -> [])
   in
-  let stream, push, close = Push_stream.make () in
-  let context : Fiber.context = { push; close; pending = 1; index = initial_index; env } in
+  let stream, context = Stream.make ~env ~debug:false ~initial_index:1 in
```
davesnx: Why the initial index starts at 1?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,178 +226,430 @@ module Model = struct
...
-  let render ?(env = `Dev) ?(debug = false) ?subscribe element =
-    let initial_chunk_id = 0 in
-    let stream, push, close = Push_stream.make () in
-    let push_chunk id chunk =
-      match chunk with
-      | Chunk_value json -> push (model_to_chunk id json)
-      | Debug_info_map json -> push (debug_info_to_chunk id json)
-      | Chunk_component_ref json -> push (client_reference_to_chunk id json)
-      | Chunk_error json -> push (error_to_chunk id json)
+module Fiber = struct
```
davesnx: Fiber got a lot of line movements that makes it a bit hard to review the diff and I only see one commit, could we maybe go step by step in person?

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: You know, I was wondering if Json is better. Like, "push Json value".

davesnx: Model.chunk is good type name btw, if you go with chunk, then the type make sense: value | component_ref | debug_info

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: To avoid confusion, I changed the `on_push` to `serialize` as you pointed out at DOJO.

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: Just solved some of those issues. I'm going to look better next week at the test cases.

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: We already talked about it in person, so I'm closing this comment.

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: Index 0 is reserved for the root chunk, that's not pushed:
```ocaml
  (* Initial index is 1 because 0 is reserved for the root chunk *)
  let stream, context = Stream.make ~initial_index:1 in
  (* ... *)
  (* We don't push the root chunk to the stream, it's added to user_scripts *)
  let root_chunk = Fiber.model_to_chunk (Value root_model) 0 in
```

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: No need anymore as I'm handling it that way:
```ocaml
let to_chunk index = Html.list [ boundary_to_chunk html index; model_to_chunk (Value model) index ] in
```

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: We don't have Fiber.chunk anymore

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: It's `to_chunk` now

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -208,8 +173,17 @@ module Model = struct
...
+    match value with
+    | Value value -> value_to_chunk id value
+    | Debug_ref debug_info -> debug_info_to_chunk id debug_info
+    | Component_ref ref -> client_reference_to_chunk id ref
+    | Error (env, error) ->
+        let error_json = make_error_json ~env ~message:error.message ~stack:error.stack ~digest:error.digest in
+        error_to_chunk id error_json
+
   (* TODO: currently it's implemented only `ReactComponentInfo` from https://github.com/facebook/react/blob/152080276c61873fdfc88db7f5856332742ddb02/packages/react-server/src/ReactFlightServer.js#L3208-L3216, it lacks other types *)
-  (* TODO: Add props, stack and parentName *)
```
davesnx, line 212: is parentName not needed?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,180 +231,484 @@ module Model = struct
...
-            Lwt.async (fun () ->
-                let%lwt value = promise in
-                let json = value_to_json value in
-                context.pending <- context.pending - 1;
-                context.push chunk_id (Chunk_value json);
-                if context.pending = 0 then context.close ();
-                Lwt.return ());
-            `String (promise_value chunk_id)
+            let promise =
+              Lwt.map to_chunk
```
pedrobslisboa: remove this pzlpzlzplz

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: The owner we have there is already the "parentName" (If we can call it like this). All the debug info belongs to the owner component (Uppercase / Async ...)

So in that case we already have the debug owner name, for now we only needs stack and props.
Even the `name` here could be changed to `ownerName` only as name on react is deprecated.

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,150 +234,148 @@ module Model = struct
...
+                is_root := false;
+                (*
+                  If it's the root element, React returns the element payload instead of a reference value. 
+                  Root is a special case: https://github.com/facebook/react/blob/f3a803617ec4ba9d14bf5205ffece28ed1496a1d/packages/react-server/src/ReactFlightServer.js#L756-L766 
+                *)
+                turn_element_into_payload element)
+              else
+                (* If it's not the root React push the element to the stream and return the reference value *)
+                let element_index =
+                  let payload = turn_element_into_payload element in
```
davesnx: maybe abstract this into a small fn?

packages/reactDom/src/ReactServerDOM.ml
davesnx: Could you undo the Fiber module changes? Meaning to not have anything other than holding context and mutable state about html?

Just to keep the diff/history alright

packages/reactDom/test/test_RSC_model.ml
```diff
@@ -370,7 +374,10 @@ let suspense_with_immediate_promise () =
   let output, subscribe = capture_stream () in
   let%lwt () = ReactServerDOM.render_model ~subscribe main in
   assert_list_of_strings !output
-    [ "0:[\"$\",\"$Sreact.suspense\",null,{\"fallback\":\"Loading...\",\"children\":\"DONE :)\"},null,[],{}]\n" ];
+    [
+      "1:\"DONE :)\"\n";
+      "0:[\"$\",\"$Sreact.suspense\",null,{\"fallback\":\"Loading...\",\"children\":\"$L1\"},null,[],{}]\n";
```
davesnx: This is conceptually incorrect. How's it possible that the resolved promise comes before the suspense node?

davesnx: even being an immediate promise, should always come after

packages/reactDom/test/test_RSC_model.ml
```diff
@@ -526,9 +536,8 @@ let client_with_element_props () =
   assert_list_of_strings !output
     [
       "1:I[\"./client-with-props.js\",[],\"ClientWithProps\"]\n";
-      "2:\"Client Content\"\n";
-      "0:[[\"$\",\"div\",null,{\"children\":\"Server \
-       Content\"},null,[],{}],[\"$\",\"$1\",null,{\"children\":\"$2\"},null,[],{}]]\n";
+      "0:[[\"$\",\"div\",null,{\"children\":\"Server Content\"},null,[],{}],[\"$\",\"$1\",null,{\"children\":\"Client \
+       Content\"},null,[],{}]]\n";
```
davesnx: What would it take to keep the model as it was before?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,150 +234,148 @@ module Model = struct
...
-              let index = use_chunk_id context in
               let message = Printexc.to_string exn in
               let stack = create_stack_trace () in
-              let error_json = make_error_json ~env:context.env ~message ~stack ~digest:"" in
-              context.push index (Chunk_error error_json);
+              let error = make_error ~message ~stack ~digest:"" in
+              let index = Stream.push ~context (to_chunk (Error (env, error))) in
               `String (lazy_value index)
-          | Return element -> turn_element_into_payload ~context element
+          | Return element ->
```
davesnx: Can you add a test case that captures this branch?

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,150 +234,148 @@ module Model = struct
...
+          turn_element_into_payload children
       | List children ->
           if is_root.contents then is_root := false;
-          `List (List.map (turn_element_into_payload ~context) children)
+          `List (List.map turn_element_into_payload children)
       | Array children ->
           if is_root.contents then is_root := false;
-          `List (Array.map (turn_element_into_payload ~context) children |> Array.to_list)
+          `List (Array.map turn_element_into_payload children |> Array.to_list)
+      (* TODO: Get the stack info from component as well *)
```
davesnx: is this comment for Array or for Uppercase?

packages/reactDom/src/ReactServerDOM.ml
davesnx: (the entire branch)

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -395,26 +386,321 @@ module Model = struct
...
+    let stream, context = Stream.make ~initial_index:0 in
+    let to_root_chunk value id =
+      let payload = client_value_to_json ~debug ~context ~to_chunk ~env value in
+      to_chunk (Value payload) id
     in
-    let context = { push = push_chunk; close; chunk_id = initial_chunk_id; pending = 0; debug; env } in
-    let initial_chunk_id = get_chunk_id context in
-    let json = client_value_to_json ~context response in
-    context.push initial_chunk_id (Chunk_value json);
+    Stream.push ~context (to_root_chunk response) |> ignore;
```
davesnx, line 415: prefer to not use ignore

packages/reactDom/src/ReactServerDOM.ml
```diff
@@ -241,150 +234,148 @@ module Model = struct
...
     | Function action ->
-        let chunk_id = use_chunk_id context in
-        context.push chunk_id (Chunk_value (`Assoc [ ("id", `String action.id); ("bound", `Null) ]));
-        `String (action_value chunk_id)
+        let index = Stream.push ~context (to_chunk (Value (`Assoc [ ("id", `String action.id); ("bound", `Null) ]))) in
+        `String (action_value index)
 
-  and client_values_to_json ~context props =
-    List.map (fun (name, value) -> (name, client_value_to_json ~context value)) props
+  and client_values_to_json ~context ~to_chunk ~env props =
```
davesnx, line 387: wdyt about calling this "client_values" and drop the "to_json"

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: Don't know, maybe I can work on a better abstraction in another PR, cause this o same "approach" is also used on Upper_case_component, It's also missing here the debug handler, and I don't want to over complicated this PR.

packages/reactDom/test/test_RSC_model.ml
```diff
@@ -409,7 +409,10 @@ let nested_suspense () =
   let output, subscribe = capture_stream () in
   let%lwt () = ReactServerDOM.render_model ~subscribe main in
   assert_list_of_strings !output
-    [ "0:[\"$\",\"$Sreact.suspense\",null,{\"fallback\":\"Loading...\",\"children\":\"$L2\"}]\n"; "1:\"DONE :)\"\n" ];
```
pedrobslisboa, line 412: I've just notice that maybe we had a bug right here. Where is the ref 2 for the `$L2`?

packages/reactDom/src/ReactServerDOM.ml
pedrobslisboa: What about `handle_client_values`?

pedrobslisboa: Keeping as it's and we're going to discuss it later.

PR #317, opened 2025-10-22
add a function to construct a url, and change how to two libs are built

EmileTrotignon:
The function works in a weird way because apparently there no way to get an empty url object in javascript.

Also no more copying files around which should improve editor integration


packages/url/js/dune
```diff
@@ -3,8 +3,7 @@
  (modes melange)
  (public_name server-reason-react.url_js)
  (libraries melange.js)
- (wrapped false)
```
pedrobslisboa, line 6: We cannot remove the wrapped false as it enables the library to function as "universal".

If you don't have it the usage becomes `Url_js.Url` instead of only `Url`, which missmatches with the native library `Url_native.Url`

packages/url/lib/dune
```diff
@@ -0,0 +1,5 @@
+(library
```
pedrobslisboa, line 1: It shouldn't work with melange as it doesn't have the universal structure. Take a look here: https://ml-in-barcelona.github.io/server-reason-react/server-reason-react/how-to-organise-universal-code.html

packages/url/lib/URL.ml
```diff
@@ -0,0 +1,7 @@
+include URL_impl
+
+let construct ?protocol ~hostname ?port ?pathname ?search ?hash ?password ?username () =
+  let apply_opt f o = match o with None -> Fun.id | Some x -> fun u -> f u x in
+  "https://example.com" |> makeExn |> apply_opt setProtocol protocol |> Fun.flip setHostname hostname
+  |> apply_opt setPort port |> apply_opt setPathname pathname |> apply_opt setSearch search |> apply_opt setHash hash
+  |> apply_opt setPassword password |> apply_opt setUsername username
```
pedrobslisboa, line 7: I'm confused about the motivation of this feature. Can you explain better in the PR description?

packages/url/native/dune
```diff
@@ -1,7 +1,7 @@
 (library
  (name url_native)
  (public_name server-reason-react.url_native)
- (wrapped false)
```
pedrobslisboa, line 4: Same as here: 
https://github.com/ml-in-barcelona/server-reason-react/pull/317/files#r2453917671

packages/url/js/dune
EmileTrotignon: Its still unwrapped because it implements `url` which is a unwrapped library. If you try to add it back you get an error message.

packages/url/js/dune
```diff
@@ -3,8 +3,7 @@
...
  (preprocess
-  (pps melange.ppx)))
+  (pps melange.ppx))
+ (implements url))
 
-(copy_files# "../URL.rei")
```
pedrobslisboa, line 10: We also don't like `copy_files`, but until we have the dune single-context feature, we need it.
(Related https://github.com/ml-in-barcelona/server-reason-react/pull/317/files#r2453920368)

PR #324, opened 2025-11-13
feature: enhance expand_styles_prop to merge className and style props

pedrobslisboa:
## Description

This PR enables the `styles` to work with both the' className' and' style' attributes.

It also fixes an error on expanding `Optional "styles"`, before it was validating the `frt` and `snd` values instead of the `styles` itself. Now, we use an `Option.map` to retrieve the correct values from the option content.
```txt
File "packages/server-reason-react-ppx/test/test.re", line 631, characters 18-24:
631 |   let div = <div ?styles />;
                        ^^^^^^
Error: This expression has type 'a option
       but an expression was expected of type 'b * 'c
Had 1 error, waiting for filesystem changes...     
```

```ocaml
(* Before *)
React.createElementWithKey ~key:None "div"
  (Stdlib.List.filter_map Stdlib.Fun.id
     [
       (match (fst x : string option) with
       | None -> None
       | Some v -> Some (React.JSX.String ("class", "className", v)));
       (match (snd x : ReactDOM.Style.t option) with
       | None -> None
       | Some v -> Some (React.JSX.Style v));
     ])
  []

(* After *)
React.createElementWithKey ~key:None "div"
  (Stdlib.List.filter_map Stdlib.Fun.id
     [
       (match (Option.map (fun x -> fst x) x : string option) with
       | None -> None
       | Some v -> Some (React.JSX.String ("class", "className", v)));
       (match (Option.map (fun x -> snd x) x : ReactDOM.Style.t option) with
       | None -> None
       | Some v -> Some (React.JSX.Style v));
     ])
  []
```

## Examples:

```reason
// Expanding styles prop into className and style props
<div styles=x />;

// Expanding styles prop into className and style props with optional styles
<div styles=?x />;

// Expanding styles prop into className and style props with className already set
<div className="lola" styles=x />;

// Expanding styles prop into className and style props with style already set
<div style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=x />;

// Expanding styles prop into className and style props with className and style already set
<div className="lola" style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=x />;

// Expanding styles prop into className and style props with optional styles prop
<div styles=?x />;

// Expanding styles prop into className and style props with optional styles prop and className already set
<div className="lola" styles=?x />;

// Expanding styles prop into className and style props with optional styles prop and style already set
<div style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=?x />;

// Expanding styles prop into className and style props with optional styles prop and className and style already set
<div className="lola" style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=?x />;```

### DIFF
--- a/packages/expand-styles-attribute/dune
+++ b/packages/expand-styles-attribute/dune
@@ -1,0 +1,6 @@
+(library
+ (name expand_styles_attribute)
+ (public_name server-reason-react.expand-styles-attribute)
+ (libraries ppxlib ppxlib.astlib)
+ (preprocess
+  (pps ppxlib.metaquot)))
--- a/packages/expand-styles-attribute/test/test.ml
+++ b/packages/expand-styles-attribute/test/test.ml
@@ -1,0 +1,87 @@
+let test_expand_styles () =
+  let loc = Ppxlib.Location.none in
+  let expr = [%expr "some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ()] in
+  let attributes = [ (Ppxlib.Labelled "styles", expr) ] in
+  let expanded_attributes = Expand_styles_attribute.make ~loc:Ppxlib.Location.none attributes in
+
+  List.iter
+    (fun attribute ->
+      match attribute with
+      | ( Ppxlib.Labelled "className",
+          [%expr fst ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ())] ) ->
+          Alcotest.(check pass) "className uses fst" () ()
+      | Ppxlib.Labelled "style", [%expr snd ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ())] ->
+          Alcotest.(check pass) "style uses snd" () ()
+      | _ -> Alcotest.fail "Expanded attributes should be className and style")
+    expanded_attributes
+
+let test_expand_styles_with_previous_className () =
+  let loc = Ppxlib.Location.none in
+  let expr = [%expr "some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ()] in
+  let attributes = [ (Ppxlib.Labelled "className", [%expr "previous-class-name"]); (Ppxlib.Labelled "styles", expr) ] in
+  let expanded_attributes = Expand_styles_attribute.make ~loc:Ppxlib.Location.none attributes in
+  List.iter
+    (fun attribute ->
+      match attribute with
+      | ( Ppxlib.Labelled "className",
+          [%expr
+            fst ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ()) ^ " " ^ "previous-class-name"]
+        ) ->
+          Alcotest.(check pass) "className uses previous class name" () ()
+      | Ppxlib.Labelled "style", [%expr snd ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ())] ->
+          Alcotest.(check pass) "style uses combine" () ()
+      | _ -> Alcotest.fail "Expanded attributes should be className and style")
+    expanded_attributes
+
+let test_expand_styles_with_previous_style () =
+  let loc = Ppxlib.Location.none in
+  let expr = [%expr "some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ()] in
+  let attributes = [ (Ppxlib.Labelled "style", [%expr "previous-style"]); (Ppxlib.Labelled "styles", expr) ] in
+  let expanded_attributes = Expand_styles_attribute.make ~loc:Ppxlib.Location.none attributes in
+  List.iter
+    (fun attribute ->
+      match attribute with
+      | ( Ppxlib.Labelled "className",
+          [%expr fst ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ())] ) ->
+          Alcotest.(check pass) "className uses fst" () ()
+      | ( Ppxlib.Labelled "style",
+          [%expr
+            ReactDOM.Style.combine "previous-style"
+              (snd ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ()))] ) ->
+          Alcotest.(check pass) "style uses combine" () ()
+      | _ -> Alcotest.fail "Expanded attributes should be className and style")
+    expanded_attributes
+
+let test_expand_styles_optional () =
+  let loc = Ppxlib.Location.none in
+  let expr = [%expr Some ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ())] in
+  let attributes = [ (Ppxlib.Optional "styles", expr) ] in
+  let expanded_attributes = Expand_styles_attribute.make ~loc:Ppxlib.Location.none attributes in
+  List.iter
+    (fun attribute ->
+      match attribute with
+      | ( Ppxlib.Optional "className",
+          [%expr
+            match Some ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ()) with
+            | None -> None
+            | Some x -> Some (fst x)] ) ->
+          Alcotest.(check pass) "className uses fst" () ()
+      | ( Ppxlib.Optional "style",
+          [%expr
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: packages/server-reason-react-ppx/cram/styles.t/input.re [hunk 1]
```diff
@@ -1 +1,26 @@
...
+// Expanding styles prop into className and style props with optional styles
+<div styles=?x />;
+
+// Expanding styles prop into className and style props with className already set
+<div className="lola" styles=x />;
+
+// Expanding styles prop into className and style props with style already set
+<div style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=x />;
+
+// Expanding styles prop into className and style props with className and style already set
```
davesnx: No need for those comments

packages/expand-styles-attribute/dune
```diff
@@ -0,0 +1,6 @@
+(library
```
davesnx, line 1: can you name the folder expand-styles as the library?

packages/expand-styles-attribute/test/test.ml
```diff
@@ -0,0 +1,216 @@
...
+      match attribute with
+      | ( Ppxlib.Labelled "className",
+          [%expr fst ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ())] ) ->
+          Alcotest.(check pass) "className uses fst" () ()
+      | ( Ppxlib.Labelled "style",
+          [%expr
+            ReactDOM.Style.combine "previous-styles"
+              (snd ("some-class-name", ReactDOM.Style.make ~backgroundColor:"gainsboro" ()))] ) ->
+          Alcotest.(check pass) "style uses combine" () ()
+      | _ -> Alcotest.fail "Expanded attributes should be className and style")
```
davesnx, line 52: why those tests don't check all properties?

packages/server-reason-react-ppx/cram/styles.t/input.re
```diff
@@ -1 +1,9 @@
+
 <div styles=x />;
+<div styles=?x />;
+<div className="lola" styles=x />;
+<div style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=x />;
+<div className="lola" style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=x />;
+<div className="lola" styles=?x />;
+<div style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=?x />;
+<div className="lola" style={ReactDOM.Style.make(~backgroundColor="gainsboro", ())} styles=?x />;
```
davesnx, line 9: I belive cram tests are good enough, no need to repeat all tests as runtime (maybe keep 1 or 2) but having repetition there seems too much, thoughts?

packages/expand-styles-attribute/test/test.ml
davesnx: anyway those tests aren't adding much here, and we could keep 1 or 2 cases and don't go full into all cases.

packages/server-reason-react-ppx/cram/component-definition.t/run.t
```diff
@@ -19,10 +19,10 @@ We need to output ML syntax here, otherwise refmt could not parse it.
             React.createElement "button"
               (Stdlib.List.filter_map Stdlib.Fun.id
                  [
-                   Some (React.JSX.Ref (buttonRef : React.domRef));
                    Some
                      (React.JSX.String
                         ("class", "className", ("FancyButton" : string)));
+                   Some (React.JSX.Ref (buttonRef : React.domRef));
                  ])
```
davesnx, line 26: any idea why this change? probably iterating right?

packages/server-reason-react-ppx/cram/styles.t/run.t
```diff
@@ -8,3 +8,128 @@ We need to output ML syntax here, otherwise refmt could not parse it.
...
+            (match x with None -> None | Some x -> Some (fst x) : string option)
+          with
+         | None -> None
+         | Some v -> Some (React.JSX.String ("class", "className", v)));
+         (match
+            (match x with None -> None | Some x -> Some (snd x)
+              : ReactDOM.Style.t option)
+          with
+         | None -> None
+         | Some v -> Some (React.JSX.Style v));
```
davesnx, line 26: this looks like unnesesary match expressions, did you check the output? Why make it like this?

pedrobslisboa: You mean, why have these two below?
1: `match x with None -> None | Some x -> Some (fst x)`
And 
2:
```ocaml
(match
  ((* $1 *) : string option)
with
| None -> None
| Some v -> Some (React.JSX.String ("class", "className", v));
```

The `2` is created by `make_prop`, so we don't have control over it, and it's not expand-styles output.
https://github.com/ml-in-barcelona/server-reason-react/blob/43ef144d3bdbe218593279ac579c2b0b0cc88d1a/packages/server-reason-react-ppx/server_reason_react_ppx.ml#L125-L129
The `1` is the one added by me, as explained in the description, we were using `fst` and `snd` on optional, which was wrong:
```ocaml
  | Ppxlib.Optional "styles" ->
      [ (Ppxlib.Optional "className", [%expr fst [%e arg]]); (Ppxlib.Optional "style", [%expr snd [%e arg]]) ]
```
```txt
File "packages/server-reason-react-ppx/test/test.re", line 631, characters 18-24:
631 |   let div = <div ?styles />;
                        ^^^^^^
Error: This expression has type 'a option
       but an expression was expected of type 'b * 'c
Had 1 error, waiting for filesystem changes...   
```

The right approach is to match the styles attribute and get the `fst` of the value. Another approach to the result is to use `Option.map` instead of `switch`, which is more elegant on the output.
```ocaml
React.createElementWithKey ~key:None "div"
  (Stdlib.List.filter_map Stdlib.Fun.id
     [
       (match (Option.map (fun x -> fst x) x : string option) with
       | None -> None
       | Some v -> Some (React.JSX.String ("class", "className", v)));
       (match (Option.map (fun x -> snd x) x : ReactDOM.Style.t option) with
       | None -> None
       | Some v -> Some (React.JSX.Style v));
     ])
  []
```

packages/server-reason-react-ppx/cram/component-definition.t/run.t
pedrobslisboa: I do, the way I'm handling the `styles` and `className/style` prop to work together, separate the className and styles from the remaining attributes, then I join them together: `([ className; style ] |> List.filter_map Stdlib.Fun.id) @ rest`

I don't like to have this change, but does the order matter in the end? I initially tried not to change the attribute order, but it led to a more complicated code than we have right now. Is it worth working on a code that doesn't change the attribute order?

packages/server-reason-react-ppx/cram/styles.t/input.re
pedrobslisboa: The point is that cram tests don't test the implementation of the code. 
For example, with the cram test, it wouldn't get this case before `<div styles=?x />`:
```
File "packages/server-reason-react-ppx/test/test.re", line 631, characters 18-24:
631 |   let div = <div ?styles />;
                        ^^^^^^
Error: This expression has type 'a option
       but an expression was expected of type 'b * 'c
Had 1 error, waiting for filesystem changes...     
```
The tests here https://github.com/ml-in-barcelona/server-reason-react/pull/324#:~:text=packages/expand_styles/test/test.ml enable us to ensure it's working as intended.

packages/server-reason-react-ppx/cram/styles.t/run.t
davesnx: okok, that makes sense. mb.

packages/server-reason-react-ppx/cram/component-definition.t/run.t
davesnx: no, just wondering if it was accidental. I can't think of a cases it matters.

packages/server-reason-react-ppx/cram/styles.t/input.re
davesnx: Yes, having 1 test like this is ok, but not all of them. That's my comment.

PR #325, opened 2025-11-15
Ensure render_model always recieve a React.element

davesnx:


pedrobslisboa:
I disagree that `render_model` should always receive a `React.element`.
To make my point, I bring some things:
1. I think we treat `render_model` as `renderToPipeableStream`, that's why we use the `arch/server/render-rsc-to-stream` to do some `model` response tests.
2. We can pass any `JSON-like` values to `React` method `renderToPipeableStream`:
	```js
	const { pipe } = renderToPipeableStream({
		element: <App />,
		params: {
			foo: "bar",
		},
		env: "development",
	}); 
	
	/** Result
	2:{"name":"App","env":"Server","key":null,"owner":null,"stack":[["","/Users/pedrolisboa/work/server-reason-react/arch/server/render-rsc-to-stream.js",51,11]],"props":{}}
	1:D"$2"
	1:["$","div",null,{"className":"container","children":["$","link",null,{"rel":"stylesheet","href":"bootstrap.min.css","precedence":"high"},"$2",[["App","/Users/pedrolisboa/work/server-reason-react/arch/server/render-rsc-to-stream.js",47,4]],1]},"$2",[],1]
	0:{"element":"$1","params":{"foo":"bar"},"env":"development"}
	*/
	```
3. The one that makes sense to only receive `React.element` is the `render_html`, which we treat similar as the renderToReadableStream:
	```js
	const { pipe } = renderToPipeableStream({
		element: <App />,
		params: {
			foo: "bar",
		},
		env: "development",
	});
	
	/** Result
	error: Objects are not valid as a React child (found: object with keys {element, bootstrapScriptContent, params, env}). If you meant to render a collection of children, use an array instead.
	*/
	```
5. We have on React a sample of `JSON-like` usage of renderToPipeableStream: https://github.com/facebook/react/blob/ea4899e13f9e29815321e3cac70fa08bb8ed790a/fixtures/flight/server/region.js#L126-L131
6. I think we should have this "complex-structure" behavior so we can send more info than only the `React.element` on cases like `navigation`

PR #329, opened 2025-12-05
Add back n forward navigation to nested router

pedrobslisboa:
## Description

This PR adds to the Nested Router demo a control over the User Agent's popstate events (back and forward)

## How

We listen for the popstate event and handle its history state:
- We get the state and use it to find the Cache on the HistoryCache (path + dynamic params are the values to find the cache)
- If the cache is found, we render the page/subpage.
- If the cache isn't found, we then 'navigate' to that page, replacing the browser history state.

## Flow
image


demo/dream-nested-router/native/shared/Router.re
```diff
@@ -24,21 +28,33 @@ module Url = {
...
-    DOM.EventTarget.dispatchEvent(
-      DOM.Event.make("popstate"),
-      DOM.Window.asEventTarget(DOM.window),
-    );
+  let replace = (state: Js.t({..}), path) => {
+    // Melange webapi don't set state type, so we use Obj.magic to cast it to the correct type
+    // https://github.com/melange-community/melange-webapi/blob/80c6ededd06cc66b75445d1ed5c855e050b156a0/src/Webapi/Dom/Webapi__Dom__History.re#L2
+    // PR: https://github.com/melange-community/melange-webapi/pull/29
+    let state: History.state = state |> Obj.magic;
+    History.replaceState(state, "", path, DOM.history);
```
davesnx, line 56: can we have a History ourselves that fixes it meanwhile, instead of the comment in 10 places? 😆

demo/dream-nested-router/native/shared/HistoryCache.re
```diff
@@ -0,0 +1,55 @@
+/**
+ * HistoryCache is a module that caches the pages.
+ * It's used to avoid fetching the same page again when navigating back and forward.
+ * For FullPage, we cache the whole page element.
+ * For SubRoute, we cache only the sub-route element.
+ */
+
```
davesnx, line 7: This looks like a FIFO cache (which is fine) but I wonder why timestamp is used for? Nothing right?

pedrobslisboa: It's FIFO, I started with timestamp and forgot to remove. Thx

PR #330, opened 2025-12-10
Add filter to esbuild plugin to scope entrypoint

pedrobslisboa:
## Description

Allow us to filter which entrypoints will receive the map.

### DIFF
--- a/demo/client/build.mjs
+++ b/demo/client/build.mjs
@@ -16,7 +16,7 @@ async function build(entryPoints, { env, output, extract, mockWebpackRequire }) 
 				target: "app",
 				mockWebpackRequire,
 				bootstrapOutput,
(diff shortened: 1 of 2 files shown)


demo/client/build.mjs
```diff
@@ -16,6 +16,7 @@ async function build(entryPoints, { env, output, extract, mockWebpackRequire })
 				target: "app",
 				mockWebpackRequire,
 				bootstrapOutput,
+				entrypointFilter: /(SinglePageRSC|DummyRouterRSC|NestedRouterRSC|ServerOnlyRSC)\.re\.js/,
```
davesnx: Does it make sense to remove "ServerOnlyRSC" from here?

also, can we have entrypointFilter be just `entrypoints` and accept: list of strings and construct the regex ourselves?

davesnx (approved):

PR #331, opened 2025-12-11
Reorganize modules of nested router

pedrobslisboa:
## Description

This PR organize better the modules of the Nested Router

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 # Changes
 
+* Add runtime error with clear message when `React.cloneElement` is used with uppercase components by @davesnx
 * Allow `[@platform js]` and `[@browser_only]` on externals to conditionally exclude them from native builds. Fixes https://github.com/ml-in-barcelona/server-reason-react/issues/170 by @davesnx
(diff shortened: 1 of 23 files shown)


.githooks/pre-push
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if ! ( make format-check ); then
+if ! ( make format ); then
```
pedrobslisboa: Revert this

demo/universal/native/shared/NestedRouter_SidebarNoteContent.re
```diff
@@ -43,17 +45,21 @@ let make =
...
+          startNavigating(() => {
+            navigate(
+              "/demo/router/"
+              ++ Int.to_string(id)
+              ++ "?"
+              ++ URL.SearchParams.toString(queryParams),
+            )
+          });
+        }
+      ]>
```
davesnx, line 62: This is a bad merge, the "?" only appears when there's queryParams. Its fixed on the left

Issue #304, opened 2025-09-17
Can we create a way to compare React and server-reason-react output?


pedrobslisboa:
## Description

As we are creating tools that replicate the server React behavior, shouldn't we have a test that compares the React and Server-reason-react outputs?

The `styled-ppx` does something similar for the murmur2 where it compares the murmur2 outputs with the ocaml implementation of it: https://github.com/davesnx/styled-ppx/blob/main/packages/murmur2/test/compare.sh

PR #352, opened 2026-02-02
compare outputs 

davesnx:
This is a draft that tries to solve https://github.com/ml-in-barcelona/server-reason-react/issues/304

I pushed the plan https://github.com/ml-in-barcelona/server-reason-react/commit/a6b2614b4d7644157671a95f9b131d32adf43424 and can be helpful to improve this setup to ensure compatibility with the React's flight format

@pedrobslisboa I know you worked on this, so this might be really helpful

### DIFF
--- a/compare/compare.sh
+++ b/compare/compare.sh
@@ -1,0 +1,194 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+usage() {
+  cat <<'EOF'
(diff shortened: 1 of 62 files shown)


davesnx:
Resolved here https://github.com/ml-in-barcelona/server-reason-react/issues/304

Issue #227, opened 2025-04-09
Add support to a more values at client_prop


pedrobslisboa:
## Description

Improve the client_props and start the movement to align the behavior of SRR RSC to what React expects.

## Why

The react expect that response of the RSC, which evolves actions response and props response to the client, to be composable.
```reason
let props = {
    bar: {
     element: <div/>
  },
  foo: "Hey",
  baz: [<div/>, "Yah!"]
}
```

But our type nowadays is:
```ocaml
and client_prop =
  (* TODO: Do we need to add more types here? *)
  | Json : Yojson.Basic.t -> client_prop
  | Element : element -> client_prop
  | Promise : 'a Js.Promise.t * ('a -> Yojson.Basic.t) -> client_prop
```

As we can see, `Json` does not support any other `client_prop` inside it, similar to `Promise` that does not support it. They always handle a `Yojson.Basic.t`.

So we cannot have: React.Promise(promise, (response) => React.Element(response)).
How would we have such response:
```reason
// Server content that needs to be serialized
let props = {
    bar: {
     element: <div> {React.String("I'm a nested element :3")} </div>
  },
  foo: [<div> {React.String("Hey")} </div>, "Yah!"],
  baz: <div> {React.String("Who is Lola?")} </div>
}

//Let's see it as client prop
{
  "bar":  React.???(["element", React.Element(<div> {React.String("I'm a nested element :3")} </div> )])
  "foo":  React.???([React.Element(<div> {React.String("Hey")} </div>, React.Json(`String("Yah!")))])
  "baz": React.Promise(React.Element(<div> {React.String("Who is Lola?")} </div>)) // This is not valid nowaday
)
```

We need to support those React.??? and we need to make Promise to accept `client_prop` instead only Yojson type.
For example (not the final code):
```ocaml
and client_prop =
  | List : client_prop list -> rsc_value
  | Assoc : (string * client_prop) list -> rsc_value
  | Json : Yojson.Basic.t -> client_prop
  | Element : element -> client_prop
  | Promise : 'a Js.Promise.t * ('a -> client_prop) -> client_prop
```

Then we can handle all those types recursively, and the final return must always be either a `Yojson.Basic.t` or a `React.element`.

```reason
{
  "bar":  React.Assoc(["element", React.Element(<div> {React.String("I'm a nested element :3")} </div> )])
  "foo":  React.List([React.Element(<div> {React.String("Hey")} </div>, React.Json(`String("Yah!")))])
[...]

PR #368, opened 2026-04-06
[@deriving rsc]

davesnx:
Solves https://github.com/ml-in-barcelona/server-reason-react/issues/227

### DIFF
--- a/packages/rsc/js/RSC.mli
+++ b/packages/rsc/js/RSC.mli
@@ -1,0 +1,49 @@
+type t
+type of_rsc_error = Rsc_error of string | Unexpected_variant of string
+
+exception Of_rsc_error of of_rsc_error
+
+val of_rsc_error_to_string : of_rsc_error -> string
+val of_rsc_error : ?depth:int -> ?width:int -> rsc:t -> string -> 'a
+val of_rsc_msg_error : string -> 'a
+val of_rsc_unexpected_variant : ?depth:int -> ?width:int -> rsc:t -> string -> 'a
+val of_rsc_msg_unexpected_variant : string -> 'a
+
+module Primitives : sig
+  val string_of_rsc : t -> string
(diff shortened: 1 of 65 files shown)


packages/rsc/ppx_js/ppx_deriving_rsc_js.ml
pedrobslisboa, line 1: Should we add a comment that this code is a "fork" of melange-json? Then we can use it as a reference for future maintenance/ideas.

packages/rsc/js/RSC.mli
```diff
@@ -0,0 +1,49 @@
...
+  val unit_of_rsc : t -> unit
+  val result_of_rsc : (t -> 'a) -> (t -> 'b) -> t -> ('a, 'b) result
+  val list_of_rsc : (t -> 'a) -> t -> 'a list
+  val array_of_rsc : (t -> 'a) -> t -> 'a array
+  val tuple2_of_rsc : (t -> 'a) -> (t -> 'b) -> t -> 'a * 'b
+  val tuple3_of_rsc : (t -> 'a) -> (t -> 'b) -> (t -> 'c) -> t -> 'a * 'b * 'c
+  val tuple4_of_rsc : (t -> 'a) -> (t -> 'b) -> (t -> 'c) -> (t -> 'd) -> t -> 'a * 'b * 'c * 'd
+  val react_element_of_rsc : t -> React.element
+  val promise_of_rsc : (t -> 'a) -> t -> 'a Js.Promise.t
+  val server_function_of_rsc : t -> 'callback Runtime.server_function
```
pedrobslisboa, line 29: Now that we have the RSC lib, what about bringing server_function to RSC?
`RSC.function` / `RSC.server_function`.

davesnx: RSC depends on React, that would cause a dependency cycle

PR #369, opened 2026-04-29
Implement let _ = xx for free variables in browser_only

davesnx:
This is an attempt to remove (or minimize a lot) the usage of 26/27/32 warnings in universal libraries.

It collects all idents used by the expression and ignores them via `let _ = xxx`. This approach was inspired by the good implementation of "collecting all idents" in https://github.com/ml-in-barcelona/react-rules-of-hooks-ppx


davesnx:
Adding `let _ = xxx` removes the warnings but it keeps the references. This makes it impossible to split code since server modules contain client refs, and vice versa.

PR #374, opened 2026-07-08
Verifiable React Flight protocol spec + full wire-format alignment with React 19.1.0

davesnx:
## What this is

A **verifiable spec of the React Flight protocol** (`packages/reactDom/react_flight_spec/`), plus fixes for **all 15 wire-format divergences it caught** — ending with srr's Flight output byte-identical to real `react-server-dom-webpack@19.1.0` across 60 covered behaviors.

Supersedes the idea behind #352, addresses #304.

## How it works

- **Single-source cases**: each `cases/shared/*.re` compiles natively (rendered by `ReactServerDOM.render_model ~env:`Prod`) *and* via melange, executed by the real react-server-dom-webpack under `bun --conditions react-server` (`generate.mjs`).
- **Committed golden fixtures** (`fixtures/*.flight`, one row per line): React's output at the exact pinned version is the ground truth in-repo.
- **Conformance suite in `dune runtest`** (offline, reads fixtures only): asserts byte-equality for every case. `~xfail=reason` records a known divergence and *asserts it still mismatches* — so a fixed divergence flips the suite loudly.
- **`make spec-generate` / `make spec-check`**: regenerate/verify fixtures against real React (needs bun; not part of `runtest`).
- A universal `Spec` harness bridges RSC constructs: native → `React.Client_component`/`Async_component`/`React.Model.*`; js → `registerClientReference` + async function components.

### Bumping React
Edit the exact pin in `spec/package.json`, `make spec-generate` — **the fixture diff is the protocol changelog**. New divergences land as xfails instead of red CI, and the conformance suite shows exactly which behaviors need work.

## Coverage (60 cases)

Elements (keys, void tags, deep nesting, large arrays), children (mixed/null/numbers), props (booleans, style objects + ordering, aria, numeric), text encoding (`$$` escaping, unicode, JSON specials), client references (dedup per module+name, nesting, kitchen-sink props), promises (`$@` rows, ordering, shared-thenable dedup, element resolution, rejection), async components (root vs nested), Suspense (symbol outlining + dedup, nesting, keys, resolution order, absent vs null fallback), errors (root/sync/async, `E`-row shape and flush order).

## The 15 divergences the spec caught — all fixed, one commit each

| Divergence | React 19.1 (prod) | srr before |
|---|---|---|
| Element tuples | `["$",type,key,props]` | 7-tuple always (now env-gated: dev keeps debug tuple) |
| `$`-string escaping | `"$$foo"` | raw `"$foo"` (ambiguous on decode) |
| Numeric props | `"tabIndex":42` | `"tabIndex":"42"` |
| Numeric children | `42`, `3.14` | `"42"`, `"3.14"` (and OCaml's `"100."`) |
| Client refs | lazy `"$L1"` | `"$1"` |
| Suspense symbol | outlined row, deduplicated | inlined `"$Sreact.suspense"` |
| Suspense fallback | absent prop omitted | `"fallback":null` |
| `aria-current` | `"aria-current"` | `"ariaCurrent"` |
| Booleanish aria | `true`/`false` | `"true"`/`"false"` |
| Style order | declaration order | reversed |
| Shared thenables | one `$@` row (writtenObjects) | duplicated rows |
| Root async components | inlined into own row | outlined `$L` |
| Root sync throw | `0:E{...}` | outlined `1:E` + `0:"$L1"` |
| E-row flush order | after model rows | before |
| (html) `hidden` attr — related, on the abort branch | | |

Structural fixes mirror React's task model: destructive root rendering (`retryTask` semantics), eager row-id allocation, deferred error-row flushing, and per-stream dedup tables (`writtenObjects`-style; thenable identity via a sealed 10-line `Physical_key` module — `Obj.repr` used for identity comparison only, never `Obj.magic`, flagged for review).

Every fix was pinned by a failing spec case *before* the encoder changed, with fresh React fixtures generated to settle behavior questions (e.g. `suspense_null_fallback` proving absent ≠ explicit-null, `promise_shared_across_components` proving dedup crosses component boundaries).

## Side effects on non-Flight output

- `{React.float(100.)}` HTML now renders `"100"` (JS formatting) instead of OCaml's `"100."`; ppx fast path fixed accordingly.
- Inline `style` HTML now emits declaration order (matches react-dom).
- Booleanish aria HTML output unchanged.

## Notes for review

- `readme.md` / `protocol.md` document the machinery, the row grammar, normalization rules (rows-not-chunks, error digests), and the bump workflow.
- The wire-format changes affect dev-mode payloads too (except the tuple gate); the browser client is React's own flight client, so decode compatibility holds by construction. Demo + full suite green.
- Known future work: reply-direction coverage (processReply → decodeReply), server functions (`$F`), hints (`H` rows), and a JS-faithful float printer for integral floats > 2^53 (pre-existing edge, documented).


davesnx:
Fixes the spec on https://github.com/ml-in-barcelona/server-reason-react/issues/304

davesnx:
Expanded the spec with three new areas (parallel workstreams, merged):

**Reply direction** (client → server): 27 cases — real `encodeReply` output from react-server-dom-webpack/client committed as fixtures, decoded by srr's `decodeReply`/`decodeFormDataReply` in conformance. Discoveries: React 19.1 encodes non-finite numbers as full words (`"$NaN"`, `"$Infinity"` — not the `$N`/`$I` short forms our hand-written decoder tests assumed), and temporary references are bare `"$T"` with no id.

**Server functions** ($F rows): 5 cases — prop, two ids, nested in object, `<form action={fn}>` — all parity except dedup (below).

**Hints** (H rows): new feature — `ReactDOM.preload`/`preconnect`/`prefetchDNS`/`preinitScript` with a request-scoped sink (Lwt.key, same pattern as the request cache), emitting id-less `:H<kind>` rows with React's flush ordering (imports → hints → model rows → errors) and per-request dedup. All 6 cases byte-equal.

Spec totals: **71 model cases + 27 reply cases**. Two new known divergences recorded as xfail (not fixed yet, kept out of the parallel workstreams on purpose):

1. **`decodeReply` `$$`-unescape bug (correctness)**: srr strips two chars instead of one — `"$$money"` (React's encoding of the user string `"$money"`) decodes to `"money"`. React's own decoder does `value.slice(1)`.
2. **Server-reference dedup**: React dedups a reference passed twice via `writtenServerReferences` (one row, both props `"$F2"`); srr emits a fresh row per occurrence.

PR #375, opened 2026-07-10
Bring react-client to the repo as submodules

pedrobslisboa:
# Build `react-client` from a vendored React submodule (drop the npm fork)

## Background

The `react-client` is not published to npm by React, it's consumed only internally by React's own bundler integrations (like react-server-dom-webpack). `ReactServerDOMEsbuild.js` needs that lib, so until now the project depended on a fork I made, `@pedrobslisboa/react-client`.

This PR removes that dependency, and instead builds `react-client` from source, from a
pinned `facebook/react` submodule, so there's no reliance on an unofficial npm package, and it's easy to update for new React versions.

## Why commit the generated /dist

Building `react-client` requires React's full Node setup, including yarn, rollup, and the submodule, which aren't available in a strictly hermetic, from-git `opam install` environment. This environment lacks `node_modules`, the submodule, and network access. By committing the /dist, we make sure that **building `server-reason-react` stays hermetic**: commands like `dune build`, `@install`, and `@demo` depend on the committed artifact and don't need Node or the submodule; only *regeneration* does.
And in the end we can have the npm package from the repo without publishing it.


pedrobslisboa:
🏓 @davesnx

PR #392, opened 2026-07-31
Add a type-safe generated router with header-negotiated navigation

davesnx:
Adds the type-safe generated router (v1). The branch started as a replacement of the nested-router prototype's `?toSubRoute=` query parameter with a header-negotiated navigation protocol, and grew into the full router: manifest, code generation, server engine, client runtime, Dream adapter, and benchmarks. The prototype is deleted.

## Manifest and code generation

Routes are declared once in a `Router.make` manifest and compiled by `server-reason-react.router-gen` (a Dune rule, not a PPX) into a universal `Router` module plus a native `RouterRegistry`:

```reason
Router.make(
  ~basePath="/demo/router",
  ~search={ text: Router.Search.optional(string) },
  ~layout=Pages.AppLayout,
  ~loading=Pages.GlobalLoading,
  ~error=Pages.AppError,
  ~notFound=Pages.NotFound,
  [
    Router.route(Home, ~page=Pages.App, ~path="/"),
    Router.group(~path="/:id<NoteId.t>", ~layout=Pages.NoteLayout, ~loader=Pages.NoteLoader,
      [Router.route(Note, ~page=Pages.Note, ~path="/")]),
  ],
);
```

- Typed path parameters (`:id<NoteId.t>` resolves `NoteId.parse`/`NoteId.to_string`), typed per-scope search parameters (`required`/`optional`/`default`/`many`).
- Every module-contract attachment takes a bare module: `~page`, `~layout`, `~loading`, `~notFound`, `~error` (policy with `status` + `make`), `~invalidSearch`, and `~loader` (implying `.load`). The generator owns the naming convention, same as `NoteId.t` codecs.
- Generated client API: typed destinations, `Router.<Name>.Link`/`href`, `useRoute` (typed active-route variant), `useNavigation`, `useSearch` (typed record + shallow updates), `useUpdateHash`.

## Navigation protocol

Navigation requests target the route URL itself and negotiate via headers. The client transmits facts, never conclusions about the shared branch:

```
GET /demo/router/42/edit
Accept:              application/react.component
Router-Navigation-From: /demo/router/42
Router-Registry:        1.<fingerprint>
Router-Base-Revision:   <revision>
```

The server matches both locations itself and computes the reusable layout prefix, requiring definition and concrete-param equality per level. Loaders mark their scope and all descendants non-reusable. Responses carry `Router-Response: full | patch | redirect | reload-required`, `Vary: Accept`, and `Cache-Control: private, no-store` on every RSC response. Every header is droppable: stripped or oversized facts degrade to a full response, never a wrong patch. A registry fingerprint from another deployment yields `reload-required` (409), applied as a hard navigation. Loader redirects stream as RSC redirect envelopes; the client validates protocol version, fingerprint, and base revision, and rejects stale responses.

## Server engine

- Loaders run sequentially, parent-first, all before streaming; children receive ancestor loader results. `Data | Error | NotFound | Redirect` results route to the nearest generated boundary.
- Per-scope `~metadata` and `~headers` resolve on every response, including patches.
- The route matcher validates and percent-decodes paths in a single pass (no `Uri.pct_decode`), with encoded-slash and dot-segment rejection. `make bench-router` benchmarks matching and parsing; CI runs it and gates regressions.

## Related renderer work included

- `ReactServerDOM.render_html` accepts `?nonce` and stamps the CSP nonce on every streamed script (payload chunks, boundary instructions, bootstrap/close scripts, bootstrap modules).
- The RSC model renderer retries components suspended by `React.Experimental.usePromise` instead of serializing the suspension as an error.
- RSC responses default `?env` to `` `Prod`` (no error messages/backtraces in payloads unless requested).

## Integration

`DreamRouter.routes` (demo) mounts the generated router: document vs Flight split, patch/redirect/reload handling, server-function POSTs, and per-request SSR control. The demo app under `demo/router` exercises the full loop.

## Validation

- Generation cycle tests: manifest -> generated code -> compiled -> engine outcomes (full/patch/redirect/reload, headers, statuses, loader ordering, boundary selection).
- Cram tests for the manifest grammar (typed destinations, reserved names, invalid paths, module-attachment errors) on OCaml >= 5.2.
- Dream adapter tests: Accept negotiation, SSR toggle, patch size, redirect envelope, registry mismatch.
- Matcher benchmarks in CI (`make bench-router`), plus the existing suites: `make format-check`, `make build`, full `make test` green on 4.14.1 and 5.4.0.

Documentation: `documentation/router.mld` and a CHANGES entry.


pedrobslisboa:
Great improve!
But can we split this PR to make it easier to review?

packages/router/native/RouterServer.ml
```diff
@@ -0,0 +1,983 @@
...
+    | DuplicateId of string
+    | DuplicatePath of string
+    | AmbiguousPattern of string * string
+    | InvalidPattern of string
+
+  let rec duplicate key = function
+    | [] -> None
+    | item :: rest ->
+        if List.exists (fun other -> String.equal (key item) (key other)) rest then Some (key item)
+        else duplicate key rest
```
pedrobslisboa, line 203: Can we rename this function and remove the key arg?
```suggestion
    let rec find_duplicated_id = function
      | [] -> None
      | item :: rest ->
          if List.exists (fun other -> String.equal id_from item.Route.id other.Route.id) rest then Some (key item.Route.id)
          else duplicate key rest
```

We can also make this function simpler without 2 loops:
```suggestion
    let find_duplicated_id routes =                                                                                                                                  
      let rec find_duplicated_id = function                                                                                                                              
        | current :: (next :: _ as rest) ->
            if String.equal current.Route.id next.Route.id then Some (DuplicateId first) else aux rest                                                                      
        | _ -> None                                                                                                                                            
      in                                                                                                                                                       
      routes |> List.sort String.compare |> find_duplicated_id
```

Question: doesn't it make more sense to get all duplicated IDs at once instead of breaking on each one?

packages/router/native/RouterServer.ml
```diff
@@ -0,0 +1,983 @@
...
+          with
+          | Some error -> Some error
+          | _ -> loop rest)
+    in
+    loop routes
+
+  let node () =
+    { static_children = Hashtbl.create 4; parameter_child = None; catch_all = None; route = None; max_specificity = -1 }
+
+  let add root route =
```
pedrobslisboa: Couldn't we make the roo the last arg so pipe works better?
```
Root |> add route 
```

packages/router/native/RouterServer.ml
```diff
@@ -0,0 +1,983 @@
...
+            let root = node () in
+            let exact = Hashtbl.create (List.length routes) in
+            List.iter
+              (fun route ->
+                add root route;
+                if route.Route.parameter_names = [] then Hashtbl.add exact route.Route.path route)
+              routes;
+            Ok { routes; exact; root })
+
+  let invalid = function
```
pedrobslisboa, line 275:
```suggestion
  let raise_invalid = function
```

pedrobslisboa (commented):
There wasn't enough time to analyze the entire structure/code. So I'm only pushing those that I've made while I was doing the review

PR #394, opened 2026-08-21
feat: make branch scope graftable, not reusable

pedrobslisboa:
## Summary

Replace the reusable-prefix calculation with insertionPoint, which compares both branches and selects the deepest shared graftable scope. This allows patches to preserve nested layouts across layout-less or loader-bearing scopes instead of falling back to the root outlet.

So when ApplyPatch is set on the context, graftAt will be the best Outlet to set the patch and `String.equal(patch.graftAt, owner)` will work to apply the page.
```
  | ApplyPatch(patch) when String.equal(patch.graftAt, owner) => (
      patch.payload,
      patch.serial > saved.serial
        ? Some({
            serial: patch.serial,
            child: patch.payload,
          })
        : None,
    )
```

## How we can see that happening (Before):
```reason
  RouterOutlet.ApplyPatch({
    serial: 42,
    graftAt: "4:root",
    targetLayouts: [
      "4:root",
      "7:group:22:id1:1",
      "14:route:EditNote2:id1:1",
    ],
    payload: editNotePayload,
  });
```

https://github.com/user-attachments/assets/057e8e3e-028d-4215-a8ee-0e7e012da13f

## After:
```reason
  RouterOutlet.ApplyPatch({
    serial: 42,
    graftAt: "7:group:22:id1:1",
    targetLayouts: [
      "4:root",
      "7:group:22:id1:1",
      "14:route:EditNote2:id1:1",
    ],
    payload: editNotePayload,
  });
```

https://github.com/user-attachments/assets/774f0b16-2b39-43be-b68e-9bfd1ef52ef9

### DIFF
--- a/demo/dream-router/test_dream_router.ml
+++ b/demo/dream-router/test_dream_router.ml
@@ -237,8 +237,8 @@ let registry_mismatch_returns_reload_required () =
   Alcotest.(check (option string)) "endpoint skipped" None !seen
 
 let patch_registry () : ((React.element, string) RouterServer.Plan.t, string) RouterServer.EndpointRegistry.t =
-  let root_branch = RouterServer.Branch.Scope.make ~id:"root" ~parameters:[] ~reusable:true in
-  let route_branch id = RouterServer.Branch.Scope.make ~id ~parameters:[] ~reusable:true in
+  let root_branch = RouterServer.Branch.Scope.make ~id:"root" ~parameters:[] ~graftable:true in
+  let route_branch id = RouterServer.Branch.Scope.make ~id ~parameters:[] ~graftable:true in
   let root_scope =
     RouterServer.Plan.Scope.make ~layout:(fun child -> React.array [| React.string (String.make 2000 'x'); child |]) ()
   in
--- a/documentation/router.mld
+++ b/documentation/router.mld
@@ -249,7 +249,7 @@ rendering begin. The scope's components receive [Data] ([~note] in the
 manifest above). [NotFound] and [Error] render the nearest boundary, and
 [Redirect(destination)] escapes the whole branch.
 
-A loader marks its scope and all descendants as non-reusable for navigation
+A loader marks its scope and all descendants as non-graftable for navigation
 patches. Keep loaders for fast validation, redirects, and route-level
 not-found decisions; stream slow data from async server components inside the
 route elements instead.
--- a/packages/router/generation/router_expansion.ml
+++ b/packages/router/generation/router_expansion.ml
@@ -679,16 +679,16 @@ let canonical_parameters ~loc parameters =
       B.pexp_tuple ~loc [ B.estring ~loc parameter.name; value ])
     parameters
 
-let branch_scope ~loc (scope : Router_declaration.scope) parameters ~reusable =
+let branch_scope ~loc (scope : Router_declaration.scope) parameters ~graftable =
   B.pexp_apply ~loc
     (B.evar ~loc "RouterServer.Branch.Scope.make")
     [
       (Labelled "id", B.estring ~loc scope.id);
       (Labelled "parameters", B.elist ~loc (canonical_parameters ~loc parameters));
-      (Labelled "reusable", B.ebool ~loc reusable);
+      (Labelled "graftable", B.ebool ~loc graftable);
     ]
 
-let scope_plan (scope : Router_declaration.scope) parameters search loaders ~reusable =
+let scope_plan (scope : Router_declaration.scope) parameters search loaders ~graftable =
   let loc = scope.loc in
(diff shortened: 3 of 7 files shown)


davesnx:
> Written by an AI coding agent (OpenCode). I hit the same bug from the other side and compared the two implementations, so posting the findings here rather than opening a competing PR.

I landed an independent fix for this on `fix/router-patch-graft-outlet` (also based on `nested-router-header-protocol`) before finding this PR. We converged on the same root cause and the same generator rule. The `test_cycle.re` hunk is byte-identical in both branches, same blob hash `5f4edea`.

The two changes are complementary. Each one fixes a real case the other gets wrong.

## The build is red because the rename is incomplete

`Scope.make` still takes `~reusable` and constructs `{ ...; reusable }` in both the `.ml` and the `.mli`, while the record field is now `graftable`:

```
File "packages/router/native/RouterServer.ml", line 514, characters 48-56:
514 |       { id; instance_key = frame id ^ identity; reusable }
                                                      ^^^^^^^^
Error: Unbound record field reusable
```

Three call-site groups already pass `~graftable`. Renaming the parameter and the field in `make`, plus line 150 of the `.mli`, makes it build and turns all 63 unit tests green.

## Comparison

I patched that locally, then ran each implementation against the other's tests.

| case | this PR | `fix/router-patch-graft-outlet` |
|---|---|---|
| loader group patch, plain group patch, revalidate, cross-instance | pass | pass |
| group loader failure clamps graft | **fail** | pass |
| group loader not-found clamps graft | **fail** | pass |
| patch through layout-less group | pass | **fail** |

### What this PR gets right and mine does not

`sharedPrefix` stops at the first non-graftable scope. `insertionPoint` keeps walking while instance keys match and remembers the deepest graftable scope. That matters when a group with no `~layout` wraps a group that has one. Mine falls back to the root, yours grafts at the inner group:

```
Expected: `"9:group:3.05:orgId1:39:projectId1:4"'
Received: `"4:root"'
```

`insertionPoint` is the better primitive and `graftable` is the better name. I kept `reusable`.

### What mine catches and this PR does not

The graft depth is computed from the branch, but the plan's scope list is only known after loaders run. A failing loader drops its own scope and everything below it from the plan, so the client no longer owns an outlet at that depth. With this PR, a failing group loader still grafts at the group:

```
Expected: `"4:root"'
Received: `"7:group:111:workspaceId1:7"'
```

The client then grafts `WorkspaceBoundary` into a `WorkspaceLayout` that a full response removes entirely, because `failure_plan` is built from `completed`, which excludes the loader-owning scope.

## Merging the two: put the bound inside the walk

My fix is `min shared (Plan.scopeCount plan)`. Dropping that onto `insertionPoint` is not safe. With `sharedPrefix`, every index below the result was graftable, so clamping always landed on a valid outlet. `insertionPoint` can return index 2 while index 1 is not graftable, and clamping to 1 selects a scope with no outlet on the client.

The bound works as a predicate instead:

```ocaml
let insertionPoint ~from ~target ~rendered =
  let rec aux index best from target =
    match (from, target) with
    | left :: from, right :: target when String.equal left.Scope.instance_key right.Scope.instance_key ->
        let best = if index < rendered && Scope.graftable right then Some (index, right) else best in
        aux (index + 1) best from target
    | _, [] -> None
    | _ -> best
  in
  aux 0 None from target
```

Called with `~rendered:(Plan.scopeCount plan)`, where `scopeCount` returns the length of the plan's scope list. It falls back to the deepest graftable scope the plan still renders, rather than clamping onto an arbitrary index.

I built this on top of `0cd4eac`: your `insertionPoint` and `graftable` naming, plus `Plan.scopeCount`, the `~rendered` bound, and my fixture and tests. 63 unit tests and 38 cycle tests pass, including all three cases that fail on one side or the other.

## Two smaller gaps

`documentation/router.mld` line 252 still says a loader marks its scope and all descendants non-reusable. This change makes that false. It is also worth documenting the tradeoff the change introduces: a reused group layout keeps the render from the previous request, so a layout built from loader data goes stale until `~revalidate=true` or a full navigation.

The only cycle-level coverage here is the one changed assertion. My branch adds a loader-less `/teams/:teamId` group and 11 cases across both group kinds, covering regular navigation, revalidate, cross-instance, and loader failure at the group and at the route. The failure cases assert an invariant rather than a literal: for the same destination, the layouts a full response renders must equal the layouts above the graft plus the layouts the patch renders.

## Happy to do the work

I can rebase my branch onto this one and push the combined version, or open a PR against this branch with just the bound, the tests, and the docs. Whichever you prefer. Your traversal should be the one that survives either way.

davesnx:
Merged. Thanks @pedrobslisboa

pedrobslisboa:
> Merged. Thanks @pedrobslisboa

Cool, I was using the comment to understand the missing flow parts I was ignoring in the review.

